PR #75, opened 2018-06-13
fix(Icon): custom svg icons in current color

SevenOutman:
Custom svg icons should apply current context's font color, for instance when used in `Nav.Item` and the nav is active.


hiyangguo:
This file is genareted by [rsuite-icon-font](https://github.com/rsuite/rsuite-icon-font) , you must add this feature in [there](https://github.com/rsuite/rsuite/blob/master/styles/less/font-awesome.less) and retry.

PR #264, opened 2018-11-15
fix TS1122: A tuple type element list cannot be empty.

zpbc007:


types/Tree.d.ts
```diff
@@ -1,9 +1,15 @@
 import * as React from 'react';
 import { TreeBaseProps } from './TreeBase';
 
+interface DataItem {
+  value: any;
+  label: React.ReactNode;
+  children?: Array<DataItem>;
```
simonguo: Array type using 'Array<T>' is forbidden for simple types. Use 'T[]' instead

https://travis-ci.org/rsuite/rsuite/jobs/455358070#L521

PR #312, opened 2018-12-18
add speaker property to WhisperProps

zpbc007:


types/Whisper.d.ts
```diff
@@ -1,5 +1,7 @@
 import * as React from 'react';
 import { AnimationEventProps, PropTypes } from '.';
+import { TooltipProps } from './Tooltip'
```
simonguo: Missing semicolon

PR #321, opened 2018-12-21
Fix some  style issues.

hiyangguo:
1.  Fixed the problem that drawer without backdrop but page unclickable.
2.  Fixed the problem that hover effect when mouse hover on  disabled picker toggle.
3.  Fixed issue where cursor is not disabled when disabled tag picker.
4.  Remove radio label & checkbox label line-height unit.

### DIFF
--- a/styles/input-picker.less
+++ b/styles/input-picker.less
@@ -42,16 +42,10 @@
   width: 100%;
 
   &-input {
-    .input-size(
-      @input-picker-toggle-content-height; @padding-base-vertical; @padding-base-input-horizontal;
-        @font-size-base; @line-height-base; @input-border-radius
-    );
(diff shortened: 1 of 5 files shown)


styles/input-picker.less
```diff
@@ -43,11 +43,15 @@
...
+      @input-picker-toggle-content-height; @padding-base-vertical; @padding-base-input-horizontal;
+        @font-size-base; @line-height-base; @input-border-radius
     );
 
     .@{tag-picker} & {
-      .input-size(@tag-picker-toggle-content-height; @padding-extra-small-vertical; @padding-base-input-horizontal; @font-size-small; @line-height-base; @input-border-radius);
+      .input-size(
+        @tag-picker-toggle-content-height; @padding-extra-small-vertical;
+          @padding-base-input-horizontal; @font-size-small; @line-height-base; @input-border-radius
+      );
```
simonguo: image

hiyangguo: Fixed.

PR #366, opened 2019-01-15
[feature] Migration momentjs to dayjs

fnpen:
Moment.js:
- big and bloated

Dayjs - 2KB **immutable** date library alternative to Moment.js with the same modern API.

Done:
- Removed moment, added dayjs
- Replaced imports moment -> dayjs
- Refactored 'sets': date.hours(2) -> date.set('hour', 2)
- Removed .clone() / Dayjs is immutable

**Results:**
rsuite.js  2.56 MiB -> rsuite.js  2.42 MiB **Δ -140KiB**
rsuite.min.js (573 KiB) -> rsuite.min.js (529 KiB) = **Δ-44KiB -7.7%**
Time: 16375ms -> Time: 8815ms **~Δ-8s**

Current state (https://bundlephobia.com/result?p=rsuite@3.5.3):
image

### DIFF
--- a/src/DateRangePicker/DateRangePicker.js
+++ b/src/DateRangePicker/DateRangePicker.js
@@ -1,7 +1,7 @@
 // @flow
 
 import * as React from 'react';
-import moment from 'moment';
+import dayjs from 'dayjs';
 import classNames from 'classnames';
 import { IntlProvider } from 'rsuite-intl';
 import _ from 'lodash';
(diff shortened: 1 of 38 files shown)


src/DateRangePicker/DateRangePicker.js
```diff
@@ -195,22 +195,19 @@ class DateRangePicker extends React.Component<Props, State> {
...
 
   // hover range presets
-  getWeekHoverRange = (date: moment$Moment) => {
+  getWeekHoverRange = (date: dayjs.Dayjs) => {
     const { isoWeek } = this.props;
 
     if (isoWeek) {
       // set to the first day of this week according to ISO 8601, 12:00 am
-      return [date.clone().startOf('isoWeek'), date.clone().endOf('isoWeek')];
+      return [date.startOf('isoWeek'), date.endOf('isoWeek')];
```
simonguo: Dayjs don't have startOf('isoWeek') at the moment

fnpen: yes, they are discussing about this...
https://github.com/iamkun/dayjs/issues/261
https://github.com/iamkun/dayjs/issues/215

I do not want patch dayjs, do we know the start of week? (sunday or monday)

fnpen:
@simonguo I look on `date-fns` also, this is nice lib and with tree-shaking.
rsuite do not have difficult functions with dates and also possible to use it.
What do you think?

simonguo:
@fnpen
Replaced with `date-fns` or `dayjs`, it is actually a good way to reduce the size of the js file. I am more worried about compatibility issues. After replacement, it will inevitably bring development costs to users.

iamkun:
I'm the author of Day.js and I'd like to help this migration if needed.

fnpen:
@

> @fnpen
> Replaced with `date-fns` or `dayjs`, it is actually a good way to reduce the size of the js file. I am more worried about compatibility issues. After replacement, it will inevitably bring development costs to users.

I think 'date-fns' is better because work with pure Date(). And component prop will take lib independent value.

I tried and this lib no so small like dayjs, but have all functions and code looks better.

fnpen:
> @fnpen
> 
> I also think `date-fns` will be better because it is pure Date. As a UI library, you should not restrict users from using `dayjs` or `momentjs`.

https://github.com/rsuite/rsuite/pull/366#issuecomment-454702965

simonguo:
> I'm the author of Day.js and I'd like to help this migration if needed.

Thanks, `dayjs` is a great lib, but it is not currently suitable for `rsuite`.

PR #476, opened 2019-05-21
Added icon names to Icon.d.ts

hiyangguo:


hiyangguo:
Generated by [rsuite-icon-font](https://github.com/rsuite/rsuite-icon-font).Template file in [here](https://github.com/rsuite/rsuite-icon-font/blob/master/src/tpl/d.ts.tpl).

PR #523, opened 2019-07-22
Implement an option for showing week numbers in the Datepickers

viart:
Covers both `<DatePicker />` & `<DateRangePicker />`

### DIFF
--- a/src/Calendar/Calendar.js
+++ b/src/Calendar/Calendar.js
@@ -35,7 +35,8 @@ type Props = {
   isoWeek?: boolean,
   limitEndYear?: number,
   className?: string,
-  classPrefix?: string
+  classPrefix?: string,
+  showWeekNumbers?: boolean
 };
 
 class Calendar extends React.PureComponent<Props> {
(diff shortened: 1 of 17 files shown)


simonguo:
Thank you very much for contributing the code, this is a great feature.

But there are two problems:
1. The style is different from what our designers demand, and there may be some minor adjustments.
2. We have now migrated from flow to typescript , which is great if you can contribute code related to typescript.
https://github.com/rsuite/rsuite/tree/typescript
https://github.com/rsuite/rsuite.github.io/tree/next

viart:
@simonguo thanks
1. it looks nice, will try to cover it
2. sure, will update and PR to those branches

viart:
@simonguo I see that in the TS version you have both TS interfaces on Props and `propTypes`, do you you really need them both? `propTypes` looks a bit redundant to me.

simonguo:
@viart 
If the developer does not use TS, propTypes is necessary as a runtime check.
However, propTypes can be removed by `babel-plugin-transform-react-remove-prop-types` when the project is built and packaged.

viart:
I see, will add them too

viart:
Closed in favour of #526

PR #526, opened 2019-07-23
Implement an option for showing week numbers in the Datepickers

viart:
The same as #523 but on top of the Typescript version.

image

Note: I can't verify `<DateRangePicker />` since the `next` version of docs even without my changes is not working for it (it's crashing and messages are pretty useless) and unfortunately I can't invest more time into debugging it right now.


simonguo:
Fix DateRangePicker module undefined problem  https://github.com/rsuite/rsuite/commit/01aa7c5aef861cd8e5185e8485fe9d99c13ef3ae

PR #591, opened 2019-09-05
use `trailWidth` and `trailColor` in `Progress.Line`

JeromeDeLeon:
I think using both props matches the `props` in `Progress.Circle` for consistency and if user did not provide `trailWidth` value then fallback to `strokeWidth` to automatically match the `strokeWidth` set by user.


simonguo:
We have now migrated from flow to typescript , which is great if you can contribute code related to typescript.

https://github.com/rsuite/rsuite/tree/next

JeromeDeLeon:
https://github.com/rsuite/rsuite/pull/597

PR #599, opened 2019-09-09
Fixed alignment of Loader within its container

Kuchasz:
Several months ago was added some piece of code that brakes center alignment of Loader within its container. Removal of that inline styles fixes that issue. 
center-is-not-in-center

### DIFF
--- a/src/Loader.js
+++ b/src/Loader.js
@@ -52,7 +52,7 @@ class Loader extends React.Component<Props> {
     return (
       <div {...props} className={classes}>
         {backdrop && <div className={addPrefix('backdrop')} />}
-        <div className={classPrefix} ref={this.loaderRef}>
+        <div className={classPrefix}>
           <span className={addPrefix('spin')} />
           {hasContent && <span className={addPrefix('content')}>{content}</span>}
         </div>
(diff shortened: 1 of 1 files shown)


src/Loader.js
```diff
@@ -24,18 +22,6 @@ class Loader extends React.Component<Props> {
...
-      addStyle(this.loader, {
-        display: isIE11() || isEdge() ? 'block' : 'table',
-        width: `${width}px`
-      });
-    }
-  }
-
   loader = null;
 
   loaderRef = ref => {
```
simonguo: `loader` and `loaderRef` also need to be deleted, declared but not used

src/Loader.js
```diff
@@ -24,24 +22,6 @@ class Loader extends React.Component<Props> {
...
-
-  loader = null;
-
-  loaderRef = ref => {
-    this.loader = ref;
-  };
-
   addPrefix(name: string) {
     return prefix(this.props.classPrefix)(name);
   }
```
simonguo: missing !
```diff
- <div className={classPrefix} ref={this.loaderRef}>
+ <div className={classPrefix}>
```

PR #667, opened 2019-10-08
Fixed modularization issue for <Calendar>.

hiyangguo:


src/Calendar/styles/themes/dark.ts
```diff
@@ -1,2 +1,3 @@
 import '../../../styles/themes/dark/core.less';
+import '../../../Button/styles/themes/dark';
```
simonguo: 少了，.less 后缀

Issue #727, opened 2019-11-14
RTL Support | Drawer Placement Problem


Anas-Abdelhadi:
### Versions

| package  | version |
| -------- | ------- |
| `react`  | `16.11` |
| `rsuite` | `4.1.1` |

### What is the expected behavior?

In the Drawer component, the direction should be **automatically** flipped without having to touch the code.

Direction does not work as expected.

### What is the current behavior?
this.toggleDrawer('left')  opens the drawer from the **left**. 
changing to **RTL** should cause the same code to open the Drawer from the **right**

### What are the steps to reproduce?
create a drawer with default settings, or specify placement as left. change the **dir** of the page to **RTL** the drawer will behave exactly similar to **LTR** case

MohammadYounes:
@simonguo The drawer component placement is fully CSS based. However, the current approach of combining both LTR/RTL styles is preventing it from operating correctly due to keyframes not being flipped (disabled [here](https://github.com/rsuite/rsuite/blob/216812af30a12dc7882470ab91039b5e64f98d12/src/styles/keyframes.less#L287-L351)).

To fix this, you have 2 options:

1. Allow the keyframes to be flipped and generate different classes 
```css
[dir=ltr] .left.slide-in {
   animation-name: slideInLeft-ltr
}

[dir=rtl] .left.slide-in {
    animation-name: slideInLeft-rtl
}
```

2. Reverse the placement in [script](https://github.com/rsuite/rsuite/blob/master/src/Drawer/Drawer.tsx#L27), which requires subscribing to `IntlContext` and swapping `left` with `right` and vice versa.

That being said,  The above options would not be required if you switch to generating a complete RTL stylesheet and the Drawer would work correctly out of the box. Moreover, you'll be able to utilize all of [rtlcss](https://rtlcss.com/) features as an added bonus.

Most likely, users will be viewing the app in either LTR or RTL language but not both, similar to light/dark themes. 

Will you be open to making the switch ?

simonguo:
@MohammadYounes 
There is some doubt about whether `right` and `left` need to be reversed.
My idea is that if the two parameters `right` and `left` are already provided to the developer, then the developer can judge according to their own locale. After all, a website will normally have only one reading habit.

Of course, I am not quite sure how this problem is understood in your country. If you really need to reverse `right` and `left`, I prefer your second option.

MohammadYounes:
> There is some doubt about whether right and left need to be reversed.

[...]

PR #733, opened 2019-11-17
feat: Reverse directional icons

MohammadYounes:


simonguo:
The icon fix for dropdown is fine.
However, as to whether it is necessary to reverse `left` and `right`, I need to think about it again.

hiyangguo:
@MohammadYounes Icon is not necessary to reverse directional .

hiyangguo:
Relative #727

MohammadYounes:
@hiyangguo Directional Icons do, for example:

* **LTR**
image

* **RTL**
image

hiyangguo:
@MohammadYounes But sometimes you use icon  `left` mean is left rather than right. So  I think icon needn't reverse, right ?

MohammadYounes:
@hiyangguo I believe having an icon that should be left in both LTR/RTL versions of the **same** app is highly unlikely. Most use cases are to indicate logical, rather than physical, direction. 

Why do you use a right pointing arrow for Next button? Because it's logical in LTR languages, where the future exists on the right. However, in RTL languages, it's the opposite, and the future exists on the left.

For example, see how Progress is affected by the language direction:

* **LTR**
image

* **RTL**
image

hiyangguo:
@MohammadYounes Maybe you are right. Thank you very much.

PR #748, opened 2019-11-22
Fix #728

### LINKED ISSUE #728: RTL Support | Tooltip Placement Problem
### Versions

| package  | version |
| -------- | ------- |
| `react`  | `16.11` |
| `rsuite` | `4.1.1` |

### What is the expected behavior?
The placement of the tooltip in **RTL** pages should be a reflection of the **LTR** pages.

### What is the current behavior?
The placement of the **tooltip does NOT** change when switching between **RTL** and **LTR**

### What are the steps to reproduce?
create a **tooltip** with default placement or set it to **Left**. switch to **RTL**. the **tooltip** still appears on the **left**.

MohammadYounes:


src/Whisper/Whisper.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 import PropTypes from 'prop-types';
 import OverlayTrigger from 'rsuite-utils/lib/Overlay/OverlayTrigger';
```
simonguo: For this PR, I prefer to put all the logic about RTL in `rsuite-utils/lib/Overlay/OverlayTrigger`.
Because, in addition to Whisper, all other Pickers use OverlayTrigger.

MohammadYounes: @simonguo All pickers use `PickerToggleTrigger` which uses `OverlayTrigger`.

The `OverlayTrigger` exists in another repo, which has no `IntlContext`. Pass direction as a prop ? Or add a context consumer to `PickerToggleTrigger` ?

```js
const PickerToggleTrigger = React.forwardRef(
  (props: PickerToggleTriggerProps, ref: React.RefObject<any>) => {
    const { pickerProps, speaker, trigger = 'click', open, ...rest } = props;
    const placement = pickerProps.placement;

    return (
      <IntlContext.Consumer>
        {context => (
          <OverlayTrigger
            trigger={trigger}
            ref={ref}
            open={open}
            placement={placementPolyfill(placement, context?.rtl)}
            speaker={speaker}
            {..._.pick(pickerProps, PickerToggleTriggerProps)}
            {...rest}
          />
        )}
      </IntlContext.Consumer>
    );
  }
);
```

simonguo: Modifying `PickerToggleTrigger` should be a more reasonable solution.
Just follow yours.

MohammadYounes: @simonguo done.

PR #749, opened 2019-11-22
Fixes #727

### LINKED ISSUE #727: RTL Support | Drawer Placement Problem

MohammadYounes:
This PR depends on https://github.com/rsuite/rsuite.github.io/pull/161

### DIFF
--- a/src/Modal/Modal.tsx
+++ b/src/Modal/Modal.tsx
@@ -8,7 +8,7 @@ import bindElementResize, { unbind as unbindElementResize } from 'element-resize
 import BaseModal from 'rsuite-utils/lib/Overlay/Modal';
 import Bounce from 'rsuite-utils/lib/Animation/Bounce';
 import { on, getHeight, isOverflowing, getScrollbarSize, ownerDocument } from 'dom-lib';
-import { prefix, defaultProps, createChainedFunction } from '../utils';
+import { prefix, defaultProps, createChainedFunction, isRTL } from '../utils';
 import ModalDialog from './ModalDialog';
 import ModalBody from './ModalBody';
 import ModalHeader from './ModalHeader';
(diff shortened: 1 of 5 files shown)


src/Modal/Modal.tsx
```diff
@@ -108,8 +108,10 @@ class Modal extends React.Component<ModalProps, ModalState> {
       bodyStyles: React.CSSProperties;
     } = {
       modalStyles: {
-        paddingRight: bodyIsOverflowing && !modalIsOverflowing ? getScrollbarSize() : 0,
-        paddingLeft: !bodyIsOverflowing && modalIsOverflowing ? getScrollbarSize() : 0
+        [doc.dir !== 'rtl' ? 'paddingRight' : 'paddingLeft']:
+          bodyIsOverflowing && !modalIsOverflowing ? getScrollbarSize() : 0,
+        [doc.dir !== 'rtl' ? 'paddingLeft' : 'paddingRight']:
+          !bodyIsOverflowing && modalIsOverflowing ? getScrollbarSize() : 0
       },
```
simonguo: In actual development, some development engineers will set the rtl property on the body tag.  So,
```js
(document.body.getAttribute('dir') || doc.dir) !=='rtl'
```
Or, add an `isRTL` method.
```js
function isRTL() {
  return (
    typeof window !== 'undefined' && (document.body.getAttribute('dir') || document.dir) === 'rtl'
  );
}
```

MohammadYounes: @simonguo The recommendation is to add `dir="rtl"` to the html tag. See: 
* https://www.w3.org/International/questions/qa-html-dir
* https://www.w3.org/International/questions/qa-bidi-css-markup 

Anyway, we can also read the effective direction:

```js
typeof window !== 'undefined' && window.getComputedStyle(document.body).direction === 'rtl';
```

Which gives the freedom of setting the direction via `dir` attribute or `direction` style on either `body` or `html` node.

PR #814, opened 2020-01-04
Display correct name for the WrappedComponent

JeromeDeLeon:
As you can see from below, I have `FlexboxGrid` but instead of the name of it, it displays `DefaultPropsComponent`. I don't know how can I affirm this one as I don't know how will I test this lib. I just saw that instead of putting name for every component, why not get it when using `defaultProps` util.

Screenshot from 2020-01-04 21-28-52

In case, it didn't work. Maybe, we need to start putting `displayName` for every component that uses `defaultProps` util

### DIFF
--- a/src/utils/defaultProps.ts
+++ b/src/utils/defaultProps.ts
@@ -10,7 +10,7 @@ function getComponentName(Component) {
   return (
     Component.displayName ||
     Component.name ||
-    typeof Component === 'string' ? Component : undefined
+    (typeof Component === 'string' ? Component : undefined)
   );
 }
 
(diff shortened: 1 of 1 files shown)


simonguo:
The `DefaultPropsComponent` was treated specially in the test case, mainly due to` renderIntoDocument`.
https://stackoverflow.com/questions/36682241/testing-functional-components-with-renderintodocument
If you want to pass the test case, you need to adjust here.
https://github.com/rsuite/rsuite/blob/dfec98f88dd3e1b0f9623b75f87fd06b0eaa0041/test/testUtils.js#L29-L33

simonguo:
@JeromeDeLeon 
Thank you for submitting the PR, this issue has been fixed.
https://github.com/rsuite/rsuite/pull/862
https://github.com/rsuite/rsuite/pull/872

PR #833, opened 2020-01-21
When the Pagination component renders a custom element, the custom element can get whether the current element is a active state property

empty916:


src/Pagination/PaginationButton.tsx
```diff
@@ -54,6 +54,7 @@ class PaginationButton extends React.Component<PaginationButtonProps> {
     const item = (
       <Component
         {...unhandled}
+        active={active}
```
simonguo: The default value of the componentClass property is SafeAnchor. SafeAnchor has no active property, which will cause a warning.
image

src/Pagination/PaginationButton.tsx
```diff
@@ -51,16 +51,29 @@ class PaginationButton extends React.Component<PaginationButtonProps> {
...
+        <Component
+          {...unhandled}
+          disabled={disabled}
+          onClick={createChainedFunction(onClick, this.handleClick)}
+        >
+          {children}
+          <Ripple />
+        </Component>
+      );
+    }
```
simonguo: Fix it as follows to make the code less redundant.
```tsx
const itemProps = {
  ...unhandled,
  disabled,
  onClick: createChainedFunction(onClick, this.handleClick)
};

if (Component !== SafeAnchor) {
  itemProps.active = active;
}

const item = (
  <Component {...itemProps}>
    {children}
    <Ripple />
  </Component>
);
```

PR #838, opened 2020-02-03
onSelect signature changed for Pagination

Zamaletdinov:


Zamaletdinov:
Hmm, not sure why, but the last time it was due timeout

simonguo:
> Hmm, not sure why, but the last time it was due timeout

This is a document deployment issue and has nothing to do with the PR you submitted.

PR #882, opened 2020-03-02
Auto change theme and resolve theme load issue.

hiyangguo:
1. Auto change theme color when toggle system dark/light mode.
2. Fixed the issue that change theme to dark , `<TopLevelNav/>` icon is not match.


hiyangguo:
@SevenOutman 😏

PR #916, opened 2020-03-20
[DateRangePicker] Update broken types and add enum for target

umens:
typing for daterange picker are inconsistent between DisabledDateFunction and DateRangePickerProps and doc


simonguo:
Thanks

PR #971, opened 2020-04-17
Italian localization

farzeni:
Hello,

Thanks for this nice library.

Here there is a proposal for Italian Localization, tell me if you think I've missed somethig.

### DIFF
--- a/src/IntlProvider/locales/it_IT.ts
+++ b/src/IntlProvider/locales/it_IT.ts
@@ -1,0 +1,66 @@
+const Calendar = {
+    sunday: 'Do',
+    monday: 'Lu',
+    tuesday: 'Ma',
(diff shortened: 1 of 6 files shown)


src/IntlProvider/locales/it_IT.ts
```diff
@@ -0,0 +1,66 @@
...
+    Uploader: {
+      inited: 'Inizializzato',
+      progress: 'Avanzamento',
+      error: 'Errore',
+      complete: 'Completato',
+      emptyFile: 'Vuoto',
+      upload: 'Carica'
+    }
+  };
+
```
simonguo, line 66: image
The code has formatting problems, you can format the code with `npm run format`.

PR #988, opened 2020-04-22
fix cant control rate and add readonly prop

nieyao:


src/Rate/Rate.tsx
```diff
@@ -25,6 +26,7 @@ export const ratePropTypes = {
   disabled: PropTypes.bool,
   max: PropTypes.number,
   renderCharacter: PropTypes.func,
+  readonly: PropTypes.bool,
```
simonguo, line 29: readonly  应该驼峰 readOnly

PR #1022, opened 2020-05-02
InputPicker - valueInSearch

AlexKletn:
valueInSearch props - saves the label as a search string.
Disabled if multi is enabled.


AlexKletn:
First of all, I work with JS, and I don’t know very well TypeSctipt features, so I could have done something wrong ..)

simonguo:
I don't quite understand what the problem solved by valueInSearch is?

If you want to add new features, you need to provide test cases, and add documentation and examples under the docs directory.

The PR failed the test and the commits log was repeated, so close it.

PR #1054, opened 2020-05-13
feat: added translations for Spanish (Argentina & Spain)

juan-manuel-alberro:
Added translations for Spanish (Argentina `es_AR` & Spain `es_ES`)

### DIFF
--- a/src/IntlProvider/locales/index.ts
+++ b/src/IntlProvider/locales/index.ts
@@ -2,6 +2,8 @@ export { default as arEG } from './ar_EG';
 export { default as daDK } from './da_DK';
 export { default as enGB } from './en_GB';
 export { default as enUS } from './en_US';
+export { default as esAR } from './es_AR';
+export { default as esES } from './es_ES';
 export { default as fiFI } from './fi_FI';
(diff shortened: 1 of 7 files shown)


src/IntlProvider/locales/index.ts
```diff
@@ -2,6 +2,8 @@ export { default as arEG } from './ar_EG';
 export { default as daDK } from './da_DK';
 export { default as enGB } from './en_GB';
 export { default as enUS } from './en_US';
+export { default as esAR } from './es_AR';
+export { default as esES } from './es_ES';
```
simonguo, line 6: @juan-manuel-alberro 
Thank you very much for the language file you contributed, but I have a question.
The contents of the two language packs es_AR and es_ES are the same, is it necessary to make both exist?

juan-manuel-alberro: @simonguo just updated the documentation!
Regarding the files, is common in the Spanish language to have multiple files since in Mexico, Colombia, etc the translation could be different. 
Here is the same but gives the flexibility to just update the words later.
I can remove the `es_ES` if consider the opposite, as you wish :)

PR #1095, opened 2020-05-29
Fix CheckPicker disabledItemValues any[][]

MarvelSQ:


MarvelSQ:
in CheckPicker disabledItemValues type should be any[], not any[][]

Issue #1107, opened 2020-06-06
<Carousel> does not auto play when children are created by array.


StrangersSoftware:
| package  | version |
| -------- | ------- |
| `react`  | `16.13.1` |
| `rsuite` | `4.7.0` |

Auto play does not start in case if children are renderend in following way:

```
<Carousel autoplay style={{width: '100%', borderRadius:5}}>
      {this.state.photo?.map((item,index) => <img key={index} alt='' src={apiUrl+apiVer+'room/getphoto/' + item}  height="450" />)}
</Carousel>
```
In case if children are hardcoded, as shown in examples, all fine.

PR #1120, opened 2020-06-10
fix: autoplay can now be interrupted in Carousel

Sleaf:
- fix: animation in `rtl` and `placement=left` or `placement=right`
- `autoplay` can be interrupted now.
- fix some error when changing children node . fix: #1107


Sleaf:
已知问题：为了`label`的`htmlFor`而使用随机生成的key充当`input`的id，在SSR应用中存在前后端渲染结果不一致的问题。

PR #1256, opened 2020-08-24
refactor(Panel): refactor Panel for RSuite v5

LeiHanCN:


src/Panel/Panel.tsx
```diff
@@ -1,189 +1,211 @@
-import * as React from 'react';
```
simonguo, line 1: 文档没有更新，参考 Message 组件
https://github.com/rsuite/rsuite/tree/next/docs/pages/components/message

src/Panel/Panel.tsx
```diff
@@ -1,189 +1,228 @@
...
+                ref={ref}
+              >
+                {renderBody()}
+              </div>
+            );
+          }}
+        </Collapse>
+      ),
+      [expanded, id, merge, prefix, renderBody]
+    );
```
simonguo, line 160: React Hook useCallback has missing dependencies: 'onEnter', 'onEntered', 'onEntering', 'onExit', 'onExited', and 'onExiting'. Either include them or remove the dependency array

Issue #1342, opened 2020-11-08
Disabled `IconButton` still show background color change on hover


rehandalal:
### Versions

| package  | version |
| -------- | ------- |
| `react`  | `16.14.0` |
| `rsuite` | `4.8.4` |

### What is the expected behavior?
When hovering the cursor over a disabled `IconButton` the background of the icon area of the button should not change color.

### What is the current behavior?
![image](https://user-images.githubusercontent.com/987136/98483061-8fd74d00-21d3-11eb-9857-9b4eedba083e.png)
The button should still look like the button on the right.

### What are the steps to reproduce?
- Create an `IconButton`:
  ```js
    <IconButton disabled icon={<Icon icon="edit" />} />
  ```
- Hover mouse over disabled `IconButton`.

CH0213:
Can I take this issue? I want to fix this bug.

CH0213:
Let me fix this bug!

simonguo:
> Let me fix this bug!

@CH0213  👌

CH0213:
I found the cause of this bug. I'll fix it soon and pull request. 😀

PR #1379, opened 2020-12-04
fix: Disabled 'IconButton' change on hover(#1342)

CH0213:
There were no conditions for disabled in the hover on [src/IconButton/common.less].

So I added a condition to the hover that if not disabled, the hover.


simonguo:
It has been fixed. 
```
npm i rsuite@4.8.6
```

PR #1535, opened 2021-02-28
fix(tree): fix `spin` icon does not display when Tree is loaded async

simonguo:


src/TreePicker/styles/mixin.less
```diff
@@ -3,4 +3,5 @@
   height: 36px;
   display: flex;
   align-items: center;
+  font-size: 14px;
```
simonguo, line 6: 字体大小的设置改为变量: @font-size-base

Issue #1567, opened 2021-03-12
onCreate prop for creatable inputPicker


BossBele:
### What problem does this feature solve?

This feature will help when trying to do something with a value not previously on the data list. Use case involves, fetching an array of data as data list and adding a created value to the database so that the user does not have to create it again.

### What does the proposed API look like?

```
<InputPicker
onCreate={(createdValue) => } />

```

BossBele:
Submitted PR #1586

PR #1586, opened 2021-03-20
Added onCreate prop for creatable inputPicker

BossBele:
This feature will help when trying to do something with a value not previously on the data list. Use case involves, fetching an array of data as data list and adding a created value to the database so that the user does not have to create it again.
Created from ISSUE #1567 

The onCreate prop is a function that fires when the `creatable` `inputPicker` has a new created value that is not in the list. The function passes values as the onSelect prop does:
`onCreate={(value, item, event) => null}`

### Example:
```
<InputPicker
creatable
onCreate={(value, item, event) => addToDatabase(value)} />
```

### DIFF
--- a/src/InputPicker/InputPicker.tsx
+++ b/src/InputPicker/InputPicker.tsx
@@ -141,6 +141,7 @@ const InputPicker: PickerComponent<InputPickerProps> = React.forwardRef(
       onExited,
       onChange,
       onClean,
+      onCreate,
       onSearch,
       onSelect,
       onOpen,
@@ -313,10 +314,11 @@ const InputPicker: PickerComponent<InputPickerProps> = React.forwardRef(
 
         if (creatable && item.create) {
           delete item.create;
+          onCreate?.(value, item, event);
           setNewData(newData.concat(item));
         }
       },
-      [creatable, newData, onSelect]
+      [creatable, newData, onSelect, onCreate]
     );
 
     /**
@@ -762,6 +764,7 @@ InputPicker.propTypes = {
   renderMenu: PropTypes.func,
   renderMenuItem: PropTypes.func,
   renderMenuGroup: PropTypes.func,
+  onCreate: PropTypes.func,
   onSelect: PropTypes.func,
   onGroupTitleClick: PropTypes.func,
   onSearch: PropTypes.func,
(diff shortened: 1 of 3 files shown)


src/InputPicker/InputPicker.tsx
```diff
@@ -313,10 +314,11 @@ const InputPicker: PickerComponent<InputPickerProps> = React.forwardRef(
 
         if (creatable && item.create) {
           delete item.create;
+          onCreate?.(value, item, event);
```
simonguo, line 317: You should add corresponding unit tests for the `onCreate` prop.

simonguo: https://github.com/rsuite/rsuite#test-driven-developmenttdd

BossBele: On it

simonguo: https://github.com/rsuite/rsuite/blob/next/src/InputPicker/test/InputPickerSpec.js

BossBele: @simonguo Added tests. Having problems with `npm install` while testing. Should I submit a bug?

src/SelectPicker/SelectPicker.tsx
```diff
@@ -83,6 +83,9 @@ export interface SelectProps<T = ValueType> {
   /** Called when the option is selected */
   onSelect?: (value: any, item: ItemDataType, event: React.SyntheticEvent) => void;
 
+  /** Called when the option is created */
+  onCreate?: (value: any, item: ItemDataType, event: React.SyntheticEvent) => void;
```
simonguo, line 87: `onCreate` is a property of InputPicker, so it should not be declared in SelectPicker.

src/InputPicker/test/InputPickerSpec.js
```diff
@@ -330,4 +330,18 @@ describe('InputPicker', () => {
...
+      if (!(data.includes(value) || data.includes(item.value))) {
+        done();
+      }
+    };
+    const instance = getInstance(
+      <InputPicker defaultOpen data={data} onCreate={doneOp} defaultValue={'Kariane'} creatable />
+    );
+
+    ReactTestUtils.Simulate.keyDown(instance.target, { keyCode: 40 });
+    ReactTestUtils.Simulate.keyDown(instance.target, { keyCode: 13 });
```
simonguo, line 345: Will not trigger `onCreate` without generating new options. Let me solve this problem

src/InputPicker/InputPicker.tsx
simonguo: There is an error in the tests, let me check.

simonguo: In addition, the documentation update instructions for onCreate are missing.

PR #1589, opened 2021-03-26
fix(pickers): Fixed the inability to pass the type check when the `labelKey` and `valueKey` of the pickers are changed.

LeiHanCN:
图片

### DIFF
--- a/src/@types/common.ts
+++ b/src/@types/common.ts
@@ -204,7 +204,7 @@ export interface SVGIcon {
   id: string;
 }
 
-export interface ItemDataType {
+export interface ItemDataType extends Record<string, any> {
   label?: string | React.ReactNode;
   value?: string | number;
   groupBy?: string;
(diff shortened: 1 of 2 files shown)


src/@types/common.ts
```diff
@@ -204,7 +204,7 @@ export interface SVGIcon {
   id: string;
 }
 
-export interface ItemDataType {
+export interface ItemDataType extends Record<string, any> {
```
simonguo, line 207:
```tsx
src/Tree/Tree.tsx:143:4 - error TS2322: Type '{ data: ItemDataType[]; value?: string | number; virtualized?: boolean; labelKey?: string | number; valueKey?: string | number; childrenKey?: string | number; defaultValue?: string | number; ... 29 more ...; ref: Ref<...>; }' is not assignable to type 'TreePickerProps<string | number>'.
  Types of property 'valueKey' are incompatible.
    Type 'string | number' is not assignable to type 'string'.
      Type 'number' is not assignable to type 'string'.

143   <TreePicker inline ref={ref} {...props} />
       ~~~~~~~~~~
```
https://ci.codesandbox.io/status/rsuite/rsuite/pr/1589/builds/113479

superman66 (approved):

Issue #1568, opened 2021-03-13
Cascader设置childrenKey后，搜索失效


ThestralZhang:
### What version of rsuite are you using?
4.9.3

### What version of React are you using?
16.13.1

### What browser are you using?
Chrome

### Describe the Bug
为Cascader指定了“children”之外的childrenKey后，搜索任意字符结果都为空

### Expected Behavior
指定其他childrenKey，搜索仍应起效

### To Reproduce
```

const myData = [{ label: 'a', value: 1, sub: [{label: 'b', value: 2}]}];
const instance = <Cascader data={myData} childrenKey="sub" />;
ReactDOM.render(instance);

```
<img width="501" alt="Screen Shot 2021-03-13 at 7 16 45 PM" src="https://user-images.githubusercontent.com/8121474/111028408-681e1d00-8431-11eb-8e7a-15b275c77e66.png">

PR #1666, opened 2021-05-13
fix(cascader): fixing childrenKey property not work when searching(#1568)

liwangping:
Fixing childrenKey property of the Cascader component does not work when searching  and Increase the corresponding unit test


src/Cascader/test/CascaderSpec.js
```diff
@@ -332,6 +332,65 @@ describe('Cascader', () => {
...
+    ReactTestUtils.act(() => {
+      const searchResult = cascaderRef.current.overlay.querySelectorAll('.rs-picker-cascader-row');
+
+      assert.equal(
+        searchResult.length,
+        2
+      );
+
+    });
+  });
```
simonguo, line 392: image

PR #1872, opened 2021-09-01
test(daterange-picker): fix flaky test scripts

SevenOutman:
Clicking the earliest few dates (first row in a month) causes the month panel to re-render. Reference of the button obtained before the re-render will become unable to click after the re-render. 

https://user-images.githubusercontent.com/8225666/131662100-2a56a338-5f28-49ce-a2e4-7148a05f8067.mov


simonguo (commented):
在另外一个 PR 中修复了https://github.com/rsuite/rsuite/pull/1862/commits/dec555eac1277bed146ad81175df35b618f70c7f

PR #1921, opened 2021-09-22
docs(v5): fix typo and add missing Form.ErrorMessage in title

maxpou:
Hi, I've found some typos when I make some experiments with your library.
Lemme know if additional stuff needs to be done.

And thanks for the job 👏


simonguo (approved):
❤️

PR #2106, opened 2021-11-05
Add button radius variable

yybawang:
Some people like right-angle buttons, such as me, so I think it’s better to customize it. Before V4, it can be customized.


SevenOutman (approved):

SevenOutman:
@yybawang LGTM. Thank you!

simonguo (approved):

PR #2158, opened 2021-11-25
fix(component): uploader fileMap is not updated when status is error

timoyan:
- ~Fix wrong parameter name in function `handleAjaxUploadError`~
- When we control the component `Uploader` state via props `fileList`, fileMap won't be updated if uploaded failed since nextFile.progress is always undefined. It cause the function `getFileList` always return incorrect info across in many caller functions.

### DIFF
--- a/src/Uploader/Uploader.tsx
+++ b/src/Uploader/Uploader.tsx
@@ -215,7 +215,7 @@ class Uploader extends React.Component<UploaderProps, UploaderState> {
 
   handleAjaxUploadError = (
     file: FileType,
-    response: object,
+    status: object,
     event: React.SyntheticEvent<any>,
     xhr: XMLHttpRequest
   ) => {
@@ -224,7 +224,7 @@ class Uploader extends React.Component<UploaderProps, UploaderState> {
       status: 'error'
     };
     this.updateFileList(nextFile, () => {
-      this.props.onError?.(response, nextFile, event, xhr);
+      this.props.onError?.(status, nextFile, event, xhr);
     });
   };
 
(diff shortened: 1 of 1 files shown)


timoyan:
I was trying to upgrade to v5 but there are such many break changes, so I raise the PR base branch v4 and hope it could be the patch version to fix this issue.

I checked the v5 source code, it is rewritten so I didn't raise the PR to v5.

SevenOutman:
Hi @timoyan, thanks for contributing this fix! I'm going to review your changes very soon.

src/Uploader/Uploader.tsx
```diff
@@ -215,7 +215,7 @@ class Uploader extends React.Component<UploaderProps, UploaderState> {
 
   handleAjaxUploadError = (
     file: FileType,
-    status: object,
+    response: object,
```
SevenOutman, line 218: This is not a mis-naming. The `handleAjaxUploaderError` function is indeed called with an `ErrorStatus` object instead of a `response` as shown here https://github.com/rsuite/rsuite/blob/v4/src/utils/ajaxUpload.ts#L14

SevenOutman (changes requested):
Hi @timoyan, the fix seems to be working well 👍. But there's a little change I'd like you to make. 👀

timoyan:
> Hi @timoyan, the fix seems to be working well 👍. But there's a little change I'd like you to make. 👀

Okay, I submit a commit to revert the rename change. Please check it, thanks.

SevenOutman:
LGTM.

SevenOutman (approved):
LGTM

timoyan:
@SevenOutman One question, when will be the recent release day in the future?

SevenOutman:
> @SevenOutman One question, when will be the recent release day in the future?

We're releasing every Friday now (both v5 and v4). So hopefully this fix will be released today.

timoyan:
> > @SevenOutman One question, when will be the recent release day in the future?
> 
> We're releasing every Friday now (both v5 and v4). So hopefully this fix will be released today.

So lucky, thanks man.

PR #2170, opened 2021-11-29
fix(Carousel): make onSelect prop optional

maxpou:
Hey, I'm trying to use the Carousel component (https://rsuitejs.com/components/carousel/) in our project but, it requires an `onSelect` props. I don't think this property needs to be required. Right?


SevenOutman:
Hi @maxpou, you're right `onSelect` should be optional. Thanks for contributing this fix!

SevenOutman (approved):

PR #2182, opened 2021-12-07
feat(FormControl): allow explicit default value

asbermudez:
### Rationale:
In the current implementation, if a user explicitly sets a default value to a `FormControl`, it never gets applied as it is overwritten by the `Form` context `defaultValues`. In some scenarios, either for convention of the project, structure/reusable components and other situations, it is desirable that the component can be set with default value regardless of the `Form` configuration.

### Solution:
We're allowing to set the `defaultValue` prop in the component and it gets used preferentially over the Form configuration if set.

### DIFF
--- a/src/FormControl/FormControl.tsx
+++ b/src/FormControl/FormControl.tsx
@@ -87,6 +87,7 @@ const FormControl: FormControlComponent = React.forwardRef((props: FormControlPr
     disabled = disabledContext,
     onChange,
     onBlur,
+    defaultValue,
     ...rest
   } = props;
 
@@ -176,7 +177,7 @@ const FormControl: FormControlComponent = React.forwardRef((props: FormControlPr
         name={name}
         onChange={handleFieldChange}
         onBlur={handleFieldBlur}
-        defaultValue={formDefaultValue[name]}
+        defaultValue={defaultValue ?? formDefaultValue[name]}
         value={val}
       />
 
--- a/src/FormControl/test/FormControlSpec.js
+++ b/src/FormControl/test/FormControlSpec.js
@@ -118,6 +118,36 @@ describe('FormControl', () => {
     assert.equal(instance.querySelector('input').value, '');
   });
 
+  it('Should render correctly form default value when set', () => {
+    const mockValue = 'value';
+    const instance = getDOMNode(
+      <Form formDefaultValue={{ name: mockValue }}>
+        <FormControl name="name" />
+      </Form>
+    );
+    assert.equal(instance.querySelector('input').value, mockValue);
+  });
+
+  it('Should render correctly default value when explicitly set and form default is not set', () => {
+    const mockValue = 'value';
+    const instance = getDOMNode(
+      <Form formDefaultValue={null}>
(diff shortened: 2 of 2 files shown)


SevenOutman:
Hi @asbermudez, could you explain what is the situation where you choose to use `<FormControl>` but don't want it's value to be controlled by `<Form>`? That way you could use `<Input>` directly IMO.

asbermudez:
> Hi @asbermudez, could you explain what is the situation where you choose to use `<FormControl>` but don't want it's value to be controlled by `<Form>`? That way you could use `<Input>` directly IMO.

@SevenOutman Well, specifically is a component that contains a part of a Form that is shared with several forms but self-contains all the logic related to the data and so on, for the form handling itself we make use of Formik. We use the form controls and so to compose the forms with labels in and errors. Because this segment is shared, it would mean that in each place where we have a form that uses it we would have to specify the defaultValue making it less maintainable and so.

SevenOutman:
> for the form handling itself we make use of Formik. We use the form controls and so to compose the forms with labels in and errors.

It's ideal to use `<FormControl>` in this use case as far as I understand. `<FormControl>` is designed to be used in conjunction with `<Form>`, where form values and errors are stored and then passed down via context. The value of of `<FormControl>` is bound to be controlled and if you want to set an initial value for it you should set it in an initial `<Form formValue>`.

 If you're using Formik and not storing form states in rsuite `<Form>`, you're not able to retrieve form errors in `<FormControl>`. I think you should use Formik `<Field>` in this case (with some wrapping over rsuite components for compatibility). WDYT?

asbermudez:
> If you're using Formik and not storing form states in rsuite `<Form>`, you're not able to retrieve form errors in `<FormControl>`. I think you should use Formik `<Field>` instead in this case (with some wrapping over rsuite components for compatibility). And also if you want to set initial value for a specific field you should set it in `initialValues` in `<Formik>` rather than on a `<Field>`.
> 
> WDYT?

Which way we control the state is not the relevant bit, the point is that as this is going to be shared through different forms, we do not want to have to duplicate the default value through and the consumer of the component will not need to know what internal value is using because it just exposes values through "onChange" and so.
In any case, conceptually, we do allow to overwrite all properties when we use it as `FormControl`, including value, but we do not allow so for the `defaultValue`. This is both confusing and, IMO, incorrect.

SevenOutman:
> In any case, conceptually, we do allow to overwrite all properties when we use it as FormControl, including value, but we do not allow so for the defaultValue. This is both confusing and, IMO, incorrect.

Fair point. But I personally don't favor explicitly controlling value on `<FormControl>` after all.

@simonguo WDYT?

myNameIsDu:
Hi @asbermudez, we've recently adjusted the handling flow of using `defaultValue` and `value` of the `Form` and the `FormControl`, we will get contrary behaviour after this fix https://github.com/rsuite/rsuite/pull/3399. 

we're particularly interested in your feedback since you implemented the feature(FormControl`'s `defaultValue` will override `Form`'s defaultValue) and your thoughts about our new design.

PR #2191, opened 2021-12-09
fix(tree): judgment error when the child node children of the tree co…

nanfeiyan123:
Judgment error when the child node children of the tree component is an empty array

image


simonguo:
Hi @nanfeiyan123  
In some cases, the data of the child node is loaded asynchronously, so keep the expandable button in the case of an empty array.  https://rsuitejs.com/components/tree/#async

I recommend you to set `children` to `null` or `undefined`.

PR #2220, opened 2021-12-15
perf(DatePicker): improve date grid a11y

SevenOutman:
### Calendar dialog

Use `dialog` role instead of `listbox`.

### Calendar

Use `grid` role instead of `table` for the month view. Associate accessible name.

### Date cells

Use `gridcell` role instead of `cell` for date cells. Add `aria-selected` state when selected.
Use `rowheader` role for week numbers.


simonguo (changes requested):
image

simonguo (approved):

Issue #2153, opened 2021-11-22
Prop `id` did not match.


welkinwong:
I'm having something Dropdown

When rendering the page in the server, I get this

```
Warning: Prop `id` did not match. Server: "menubutton-11" Client: "menubutton-3"
button
Button@http://localhost:3000/packages/modules.js?hash=d3bc85f5d0c31f910b588cfada6e48734864f8e2:40390:17
IconButton@http://localhost:3000/packages/modules.js?hash=d3bc85f5d0c31f910b588cfada6e48734864f8e2:41775:19
DropdownToggle@http://localhost:3000/packages/modules.js?hash=d3bc85f5d0c31f910b588cfada6e48734864f8e2:42639:24
div
Menu@http://localhost:3000/packages/modules.js?hash=d3bc85f5d0c31f910b588cfada6e48734864f8e2:42715:23
Dropdown@http://localhost:3000/packages/modules.js?hash=d3bc85f5d0c31f910b588cfada6e48734864f8e2:41859:24
div
styled.div@http://localhost:3000/packages/modules.js?hash=d3bc85f5d0c31f910b588cfada6e48734864f8e2:106206:19674
header
styled.header@http://localhost:3000/packages/modules.js?hash=d3bc85f5d0c31f910b588cfada6e48734864f8e2:106206:19674
Header@http://localhost:3000/app/app.js?hash=7809789ca0509f6ddf0add4cc64affde03ac5cbd:2299:13
Home
RequireAuth@http://localhost:3000/app/app.js?hash=7809789ca0509f6ddf0add4cc64affde03ac5cbd:3676:15
Routes@http://localhost:3000/packages/modules.js?hash=d3bc85f5d0c31f910b588cfada6e48734864f8e2:32773:12
App
HelmetProvider@http://localhost:3000/packages/modules.js?hash=d3bc85f5d0c31f910b588cfada6e48734864f8e2:105861:6594
CustomProvider@http://localhost:3000/packages/modules.js?hash=d3bc85f5d0c31f910b588cfada6e48734864f8e2:33533:23
AppProvider@http://localhost:3000/app/app.js?hash=7809789ca0509f6ddf0add4cc64affde03ac5cbd:3786:28
Router@http://localhost:3000/packages/modules.js?hash=d3bc85f5d0c31f910b588cfada6e48734864f8e2:32711:12
[...]

PR #2239, opened 2021-12-17
fix(useUniqueId): fix issue #2153

crossworth:
Hello, I think this PR fixes the #2153 issue.

I have changed the `useUniqueId` to return a empty string and (with a `useEffect`) change it to a real unique id.

This way both the server and client will have the same output when using SSR.

I'm not sure if this is the best approach, but it works, I have tested with a local project and now the error is gone.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -3,7 +3,7 @@
  * Run all tests: `npm run tdd`
  * Run all styles tests: `M=styles npm run tdd`
  * Run a component test: `M=Button npm run tdd`
- * Run a test of a file: `src/Picker/test/PickerToggleSpec.js npm run tdd`
+ * Run a test of a file: `F=src/Picker/test/PickerToggleSpec.js npm run tdd`
  */
 
 /**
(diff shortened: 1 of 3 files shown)


SevenOutman:
Hi @crossworth, thanks for contributing this edit! But does this implementation cause redundant re-renders in client-only scenarios?

crossworth:
Hello @SevenOutman I think it will, but I'm not sure if we can avoid it.

When using SSR we are required to have the same value rendered on the server match the value rendered on the client (for the hydration process).

I dont think we can generate an id by a deterministic way.

So the options we have is to replace the id after (this PR) or manually provide it.

Maybe we could set up some how if we want this feature enable?

SevenOutman:
> I dont think we can generate an id by a deterministic way.

What about checking `window` existence and change the initial value of `idState` accordingly?

crossworth:
Sorry for the delay, that was my first thought as well, but it will not work.

The root problem is that the server MUST return the same output as the client when using SSR.

If we check the `window` on the server it will return an empty id first (this PR), than replace with a correct id, but on the client it will have correct id from the start (and possible a different one).

Searching a bit more about this problem I have found this issue:
https://github.com/facebook/react/issues/5867

Maybe we should switch to https://github.com/thearnica/react-uid?

But from what I have seen all the solutions use some kind of state.

PR #2244, opened 2021-12-20
perf(Context): use `useMemo` to memozie the values returned to the Context Provider

simonguo:


src/Calendar/Calendar.tsx
```diff
@@ -167,23 +167,43 @@ const Calendar: RsRefForwardingComponent<'div', CalendarProps> = React.forwardRe
...
+      }),
+      [
+        calendarDate,
+        dateRange,
+        format,
+        hoverRangeValue,
+        inSameMonth,
+        inSameThisMonthDate,
+        inline,
+        isDisabledDate,
```
SevenOutman, line 196: The 'isDisabledDate' function makes the dependencies of useMemo Hook (at line 188) change on every render. To fix this, wrap the definition of 'isDisabledDate' in its own useCallback() Hook

SevenOutman (changes requested):

SevenOutman (approved):

PR #2302, opened 2022-01-13
fix(Carousel): add a flex-wrap property on toolbar buttons

maxpou:
Hey, I spotted a small CSS issue with the carousel.

**Problem:**
When you have many images, some buttons can be unreachable.
We have this problem in our application. The carousel is in a card that takes 30% of the screen width. When the carousel has a "bar shape", we can't have more than 8 images.

Screenshot 2022-01-13 at 14 33 43

**Solution**: 
with this CSS property, it's automatically nicer.
Screenshot 2022-01-13 at 14 34 45

**How to reproduce:**
1. Open `docs/pages/components/carousel/fragments/basic.md`
2. Duplicate the image many times.

### DIFF
--- a/src/Carousel/styles/index.less
+++ b/src/Carousel/styles/index.less
@@ -53,6 +53,7 @@
 
     > ul {
       display: flex;
+      flex-wrap: wrap;
       list-style: none;
       margin: 0;
       padding: 0;


SevenOutman (approved):
LGTM. Thanks!

PR #2338, opened 2022-02-04
feat(Carousel): Expose active index

asbermudez:
**Changes:**
This PR exposes the active index to the user so he can either initialize with a different index or to change it externally. It keeps all the behaviors already in place.

**Rationale:**
Sometimes the user want to initialize a different initial position, for example loading last status of his navigation, or change the active element externally, for example having an expanded view of a thumbnail that has navigation between elements and needs to keep the carousel in sync.

### DIFF
--- a/docs/pages/components/carousel/fragments/position.md
+++ b/docs/pages/components/carousel/fragments/position.md
@@ -1,0 +1,17 @@
+<!--start-code-->
+
+```js
+const instance = (
+  <Carousel index={2} className="custom-slider">
+    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=1" height="250" />
+    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=2" height="250" />
+    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=3" height="250" />
+    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=4" height="250" />
+    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=5" height="250" />
+  </Carousel>
+);
+
+ReactDOM.render(instance);
+```
+
+<!--end-code-->
--- a/docs/pages/components/carousel/fragments/position.md
+++ b/docs/pages/components/carousel/fragments/position.md
@@ -2,7 +2,7 @@
 
 ```js
 const instance = (
-  <Carousel index={2} className="custom-slider">
+  <Carousel activeIndex={2} className="custom-slider">
     <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=1" height="250" />
     <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=2" height="250" />
     <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=3" height="250" />
--- a/docs/pages/components/carousel/fragments/position.md
+++ b/docs/pages/components/carousel/fragments/position.md
@@ -1,17 +1,26 @@
 <!--start-code-->
 
 ```js
-const instance = (
-  <Carousel activeIndex={2} className="custom-slider">
-    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=1" height="250" />
-    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=2" height="250" />
-    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=3" height="250" />
-    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=4" height="250" />
-    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=5" height="250" />
-  </Carousel>
-);
+const App = () => {
+  const [activeIndex, setActiveIndex] = React.useState(2);
+  return (
+    <Carousel
+      className="custom-slider"
+      activeIndex={activeIndex}
+      onSelect={index => {
+        setActiveIndex(index);
+      }}
+    >
+      <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=1" height="250" />
+      <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=2" height="250" />
+      <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=3" height="250" />
+      <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=4" height="250" />
+      <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=5" height="250" />
+    </Carousel>
+  );
+};
 
-ReactDOM.render(instance);
+ReactDOM.render(<App />);
 ```
 
 <!--end-code-->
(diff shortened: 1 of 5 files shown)


simonguo (commented):
@asbermudez  thanks
It's a nice feature. But in the design of props we want uniformity and extensibility.

Referring to [PanelGroup](https://github.com/rsuite/rsuite/blob/main/src/PanelGroup/PanelGroup.tsx#L12) , `index` can be replaced by adding `activeIndex` and `defaultActiveIndex`.

wdyt?

asbermudez:
@simonguo Sounds good to change to `activeIndex`, I will have to rename the internal state variable though to avoid collision. I don't think makes much sense to have a `defaultActiveIndex`, as from the behavior perspective adds not much as just having an `activeIndex` that doesn't change. Should we go with `activeIndex` only?

SevenOutman:
> I don't think makes much sense to have a defaultActiveIndex, as from the behavior perspective adds not much as just having an activeIndex that doesn't change.

These two props differs on whether they are _controlled_. For instance, when controlled `activeIndex` is provided, the current slide will remain unless `activeIndex` changes, even if user clicks on its indicator or `autoplay` is on. Meanwhile the uncontrolled `defaultActiveIndex` only sets the initial slide, but don't interfere the autoplay behavior, just like `value`/`defaultValue` of an `<input>`.

docs/pages/components/carousel/zh-CN/index.md
```diff
@@ -12,6 +12,10 @@
 
 <!--{include:`basic.md`}-->
 
+### 受控指数
```
simonguo, line 15:
```suggestion
### 受控的幻灯片
```

src/Carousel/Carousel.tsx
```diff
@@ -38,6 +41,7 @@ const Carousel: RsRefForwardingComponent<'div', CarouselProps> = React.forwardRe
       shape = 'dot',
       autoplay,
       autoplayInterval = 4000,
+      activeIndex = 0,
```
simonguo, line 44:
```suggestion
      activeIndex: activeIndexProp,
```

src/Carousel/Carousel.tsx
```diff
@@ -51,26 +55,38 @@ const Carousel: RsRefForwardingComponent<'div', CarouselProps> = React.forwardRe
     const vertical = placement === 'left' || placement === 'right';
     const lengthKey = vertical ? 'height' : 'width';
 
-    const [activeIndex, setActiveIndex] = useState(0);
+    const [currentIndex, setCurrentIndex] = useState(activeIndex);
```
simonguo, line 58: Reference PanelGroup https://github.com/rsuite/rsuite/blob/main/src/PanelGroup/PanelGroup.tsx#L49
```suggestion
    const [activeIndex, setActiveIndex] = useControlled(activeIndexProp, defaultActiveIndex);
```

simonguo (changes requested):
activeIndex as a controlled property,  controlled by: onSelect, initial prop: defaultActiveindex

src/Carousel/Carousel.tsx
```diff
@@ -80,14 +96,19 @@ const Carousel: RsRefForwardingComponent<'div', CarouselProps> = React.forwardRe
...
-    // `autoplay` needs to be cast to boolean type to avoid undefined parameters.
-    const { clear, reset } = useTimeout(handleSlide, autoplayInterval, !!autoplay && count > 1);
+    useEffect(() => {
+      if (inputActiveIndex !== activeIndex) {
+        setInputActiveIndex(activeIndex);
+        if (activeIndex >= 0 && activeIndex !== currentIndex) {
+          handleSlide(activeIndex);
+        }
+      }
+    }, [activeIndex, inputActiveIndex, currentIndex, handleSlide, setInputActiveIndex]);
```
simonguo, line 111: After using useControlled to manage activeIndex and defaultActiveIndex, there is no need to update activeIndex through useEffect.
```suggestion
```

simonguo (changes requested):

asbermudez:
@simonguo @SevenOutman Done the changes as discussed :)
https://github.com/rsuite/rsuite/pull/2338/commits/4e28d4c3fe28dd2027e5e5d4599e955499f05996

src/Carousel/Carousel.tsx
asbermudez: Changed here https://github.com/rsuite/rsuite/pull/2338/commits/4e28d4c3fe28dd2027e5e5d4599e955499f05996

docs/pages/components/carousel/fragments/position.md
```diff
@@ -0,0 +1,17 @@
...
+  <Carousel activeIndex={2} className="custom-slider">
+    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=1" height="250" />
+    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=2" height="250" />
+    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=3" height="250" />
+    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=4" height="250" />
+    <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=5" height="250" />
+  </Carousel>
+);
+
+ReactDOM.render(instance);
```
simonguo, line 14:
```suggestion
const App = () => {
  const [activeIndex, setActiveIndex] = React.useState(2);
  return (
    <Carousel
      className="custom-slider"
      activeIndex={activeIndex}
      onSelect={index => {
        setActiveIndex(index);
      }}
    >
      <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=1" height="250" />
      <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=2" height="250" />
      <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=3" height="250" />
      <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=4" height="250" />
      <img src="https://via.placeholder.com/600x250/8f8e94/FFFFFF?text=5" height="250" />
    </Carousel>
  );
};

ReactDOM.render(<App />);
```

src/Carousel/test/CarouselSpec.js
```diff
@@ -91,4 +91,17 @@ describe('Carousel', () => {
...
+      <Carousel defaultActiveIndex={2}>
+        <div>1</div>
+        <div>2</div>
+        <div>3</div>
+        <div>4</div>
+      </Carousel>
+    );
+
+    assert.equal(instance.querySelector('[aria-hidden=false]').textContent, '3');
+  });
```
simonguo, line 106: Add a test case for `activeIndex`

simonguo (changes requested):

src/Carousel/test/CarouselSpec.js
asbermudez: @simonguo Could you point to a existing test case of a similar scenario? I mean a scenario in which we change the component props externally in the test. 
I've looked through the code and I couldn't find one (even the Panel component doesn't have it) and I am not sure how to implement it using this testing framework.

simonguo: reference test: https://github.com/rsuite/rsuite/blob/main/src/Modal/test/ModalSpec.js#L119-L139

asbermudez: Thanks a lot! Test added here https://github.com/rsuite/rsuite/pull/2338/commits/a84e0bfda58ae3a89b127d99ee957e6d70bef966

simonguo (approved):

PR #2343, opened 2022-02-09
fix(IconButton): infer addtional props from as prop

SevenOutman:
image

### DIFF
--- a/src/IconButton/IconButton.tsx
+++ b/src/IconButton/IconButton.tsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import Button, { ButtonProps } from '../Button';
-import { useClassNames } from '../utils';
 import { IconProps } from '@rsuite/icons/lib/Icon';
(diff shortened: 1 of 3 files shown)


src/IconButton/IconButton.tsx
```diff
@@ -15,7 +16,12 @@ export interface IconButtonProps extends ButtonProps {
   placement?: 'left' | 'right';
 }
 
-const IconButton = React.forwardRef((props: IconButtonProps, ref) => {
+const IconButton: RsRefForwardingComponent<
+  typeof Button,
+  IconButtonProps & {
+    ref?: React.Ref<HTMLElement>;
```
simonguo, line 22: image

simonguo (changes requested):

src/IconButton/IconButton.tsx
SevenOutman: 已修改 LanguageButton 中的类型声明

simonguo (approved):

Issue #2336, opened 2022-01-30
Toaster push is showing unexpected behaviour when called from useEffect


utkarsh-alphonso:
### What version of rsuite are you using?

5.5.2

### What version of React are you using?

17.0.0

### What browser are you using?

Chrome

### Describe the Bug

In the useEffect of my form component, I trigger a notification for successful/unsuccessful submit. With rsuite v4, I had it working fine. But as soon as I migrated to v5, replacing the earlier Notification format with the one using toaster - I am encountering a bug as follows: 

`Uncaught TypeError: Cannot read properties of null (reading 'push')
`
which points to the line   

`return container.current.push(message) 
`
(inside the toaster.push function)

Some other observations: If I trigger the notification from a button callback - it works. Infact once I trigger it from a callback, even the useEffect triggers work fine. Please help!

simonguo:
hi @utkarsh-alphonso , can you reproduce your problem with [codesandbox](https://codesandbox.io/s/rsuite-template-5vq6zo2z5l)?

maxpou:
@simonguo I have the same issue. Any workaround?

https://codesandbox.io/s/festive-wozniak-vk379?file=/src/index.js

SevenOutman:
Closed via #2353

PR #2353, opened 2022-02-14
fix(Toaster): code breaks when toaster is in a useEffect

maxpou:
**Problem**: see #2336.

**Root cause**: during the TS -> JS compilation, the non-null assertion operator got translated from `container.current!.push` to `container.current.push`. It crash when `container.current` is null.

**How to test**:

1. open docs/pages/components/notification/index.tsx
2. replace the content with the following:

<details>
  <summary>(Click to expand)</summary>

  ```ts
import React, { useEffect } from 'react';
import { Button, Notification, toaster } from 'rsuite';

export default function Page() {
  function notify() {
    toaster.push(<Notification>Hi toaster!</Notification>);
  }

  useEffect(() => {
    notify();
  }, []);

  return (
    <div>
      Hello
      <Button onClick={notify}>React Suite</Button>
    </div>
  );
}
```
</details>

### DIFF
--- a/src/toaster/test/toasterSpec.js
+++ b/src/toaster/test/toasterSpec.js
@@ -1,5 +1,5 @@
-import React from 'react';
-import { screen } from '@testing-library/react';
+import React, { useEffect } from 'react';
+import { screen, render } from '@testing-library/react';
 import toaster from '../toaster';
 
 const element = document.createElement('div');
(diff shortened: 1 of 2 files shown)


simonguo (commented):
Hi @maxpou , thanks for fixing that. Hope you can add a test for that fix.

Reference test: https://github.com/rsuite/rsuite/blob/main/src/MaskedInput/test/TextMaskSpec.js#L27-L36

maxpou:
@simonguo I just added a test. Let me know what you think about it!

simonguo (approved):

PR #2397, opened 2022-03-13
Group By Using Dot Notation

hassanzohdy:
Hello,

This PR allow users when using `SelectInput` or `CheckPicker` to use **groupBy** prop using `dot notation` 

Example Of Usage:

```tsx
const options = [
  {
    label: 'Cairo',
    value: '1',
    country: {
      id: '1',
      name: 'Egypt'
    }
  },
  {
    label: 'Alexandria',
    value: '2',
    country: {
      id: '1',
      name: 'Egypt'
    }
  },
  {
    label: 'New York',
    value: '3',
    country: {
      id: '2',
      name: 'USA'
    }
  },
  {
    label: 'Washington',
    value: '4',
    country: {
      id: '2',
      name: 'USA'
    }
  }
];

export default function HomePage() {
  return (
    <>
      <SelectPicker placeholder="City" groupBy="country.name" data={options} />
    </>
  );
}
```


src/utils/getDataGroupBy.ts
```diff
@@ -1,3 +1,4 @@
+import { Obj } from '@mongez/reinforcements';
```
SevenOutman, line 1: You may not have to involve this package as se already have `lodash` as a dependency and you can use its `get` method.

SevenOutman (changes requested):
Hi @hassanzohdy, thanks for this edit! Here's something I'd like to suggest you change, regarding our [contribution guidelines](https://github.com/rsuite/rsuite/blob/main/CONTRIBUTING.md).

- The `package-lock.json` should not be upgraded to `lockfileVersion: 2`
- Changes in codebase should be covered with tests

I will review this PR again after these changes are applied

src/utils/getDataGroupBy.ts
hassanzohdy: @SevenOutman 

I removed the package and updated the package-lock.json file

Have a check on it now.

Regards,

hassanzohdy:
@SevenOutman

I removed the package and updated the package-lock.json file

Have a check on it now.

Regards,

SevenOutman (commented):
Could you add test cases for this `getDataGroupBy` function to cover dot-notation use cases?

hassanzohdy:
I Updated `utils/test/getDataGroupBySpec.js` file with the test case for the dot notation syntax.

SevenOutman:
> I Updated `utils/test/getDataGroupBySpec.js` file with the test case for the dot notation syntax.

The test failed ([see logs](https://github.com/rsuite/rsuite/runs/5547134555?check_suite_focus=true#step:7:2740)). Please check whether the test case or implementation is incorrect.

hassanzohdy:
I've updated the test results now it should be working just fine.

If you could tell me how i would run the test as when i run `npm run test` it just go through all tests without any results output.

SevenOutman:
The tests are passing. Congrats! However, the overall coverage has decreased due to the line addition in `src/MaskedInput/createTextMaskInputElement.ts`, could you make sure this change is necessary?

> If you could tell me how i would run the test as when i run `npm run test` it just go through all tests

Setting environment variable `F` to path of a test file make it possible to only run that specific file. And in addition to `npm run test` which runs tests once, `npm run tdd` runs the tests and watches changes in test file and source file and run the tests again automatically.

    F=src/utils/test/getDataGroupBySpec.js npm run tdd

> without any results output.

It's required to have Chrome installed to be able to run the tests locally, but I'm not sure it's relavant.

hassanzohdy:
Thanks for your feedback,

I just reverted it, it was just  typescript compile hint error, i reverted it back to the original

SevenOutman (approved):
LGTM now :)

PR #2411, opened 2022-03-22
fix: fix selectPicker when search is controlled and exit Dropdown wit…

myNameIsDu:
fix selectPicker when search is controlled and exit Dropdown without resetting external search

### DIFF
--- a/src/SelectPicker/SelectPicker.tsx
+++ b/src/SelectPicker/SelectPicker.tsx
@@ -290,7 +290,7 @@ const SelectPicker = React.forwardRef(
       setActive(false);
       onSearch?.('');
       onClose?.();
-    }, [onClose, setSearchKeyword]);
+    }, [onClose, setSearchKeyword, onSearch]);
 
     const handleEntered = useCallback(() => {
       setActive(true);
--- a/src/SelectPicker/SelectPicker.tsx
+++ b/src/SelectPicker/SelectPicker.tsx
@@ -83,7 +83,7 @@ export interface SelectProps<T> {
   onGroupTitleClick?: (event: React.SyntheticEvent) => void;
 
   /** Called when searching */
-  onSearch?: (searchKeyword: string, event: React.SyntheticEvent) => void;
+  onSearch?: (searchKeyword: string, event?: React.SyntheticEvent) => void;
 
   /** Called when clean */
   onClean?: (event: React.SyntheticEvent) => void;
@@ -288,6 +288,7 @@ const SelectPicker = React.forwardRef(
     const handleExited = useCallback(() => {
       setSearchKeyword('');
       setActive(false);
+      onSearch?.('');
       onClose?.();
     }, [onClose, setSearchKeyword]);
 
--- a/src/SelectPicker/test/SelectPickerSpec.js
+++ b/src/SelectPicker/test/SelectPickerSpec.js
@@ -400,7 +400,7 @@ describe('SelectPicker', () => {
       expect(getByTestId('content')).to.have.text('Not selected');
     });
   });
-  it('SearchWord should be reset when controlled and triggered off', async () => {
+  it('Should call onSearch when closed', async () => {
     const handleSearch = sinon.spy();
     const Wrapper = () => {
       return (
--- a/src/SelectPicker/test/SelectPickerSpec.js
+++ b/src/SelectPicker/test/SelectPickerSpec.js
@@ -1,5 +1,5 @@
-import React from 'react';
-import { render } from '@testing-library/react';
+import React, { useRef, useState } from 'react';
+import { render, cleanup, fireEvent } from '@testing-library/react';
 import ReactTestUtils from 'react-dom/test-utils';
 import { getDOMNode, getInstance } from '@test/testUtils';
 import SelectPicker from '../SelectPicker';
@@ -23,6 +23,10 @@ const data = [
     role: 'Master'
   }
 ];
+
+afterEach(() => {
+  cleanup();
+});
 
 describe('SelectPicker', () => {
   it('Should clean selected default value', () => {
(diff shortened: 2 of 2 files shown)


src/SelectPicker/SelectPicker.tsx
```diff
@@ -288,6 +288,7 @@ const SelectPicker = React.forwardRef(
     const handleExited = useCallback(() => {
       setSearchKeyword('');
       setActive(false);
+      onSearch?.('');
       onClose?.();
     }, [onClose, setSearchKeyword]);
```
simonguo, line 293:
```
React Hook useCallback has a missing dependency: 'onSearch'. Either include it or remove the dependency array
```

simonguo (changes requested):

src/SelectPicker/SelectPicker.tsx
myNameIsDu: Yeah,I added it.

simonguo (approved):

src/SelectPicker/test/SelectPickerSpec.js
```diff
@@ -400,4 +404,53 @@ describe('SelectPicker', () => {
...
+    assert.equal(searchRef, 'a');
+
+    // close select
+    fireEvent.mouseDown(exit, { bubbles: true });
+
+    promise.then(() => {
+      assert.equal(searchRef, '');
+      done();
+    });
+  });
```
SevenOutman, line 455: Please refer to existing test scripts and see how to use `sinon.spy()` to validate a callback has been called with correct arguments.

SevenOutman (changes requested):

src/SelectPicker/test/SelectPickerSpec.js
```diff
@@ -24,6 +24,10 @@ const data = [
   }
 ];
 
+afterEach(() => {
+  cleanup();
+});
+
```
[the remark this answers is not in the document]
myNameIsDu: Yeah,I removed it

src/SelectPicker/test/SelectPickerSpec.js
myNameIsDu: I modified it, but I think this test code does not reflect its operation scenario, How do you think

myNameIsDu: I mean the way this is written should not be suitable for such a description, should is "shoule call onSearch when closed", or other

SevenOutman: > I mean the way this is written should not be suitable for such a description, should is "shoule call onSearch when closed", or other

The description was not quite suitable in the first place. "Should call onSearch with '' when closed" is exactly what should be tested here. The goal of unit tests is to make sure the APIs we provide (`onSearch`) work as expected (called with '' when closed), rather than to cover how users are using our APIs (e.g. updating a state when onSearch is called).

src/SelectPicker/test/SelectPickerSpec.js
```diff
@@ -400,4 +400,24 @@ describe('SelectPicker', () => {
...
+    };
+    Wrapper.displayName = 'WrapperSelectPicker';
+    let { container } = render(<Wrapper />);
+    const exit = container.querySelector('#exit');
+
+    // close select
+    fireEvent.mouseDown(exit, { bubbles: true });
+    await waitForElementToBeRemoved(document.querySelector('.rs-picker-search-bar-input'));
+
+    assert.isTrue(handleSearch.calledOnce);
```
SevenOutman, line 421: Avoid mentioning implementation detail in test cases. Simply use `waitFor` from `@testing-library/react` for async assertions.

---

Arguments of onSearch call should also be asserted.

src/SelectPicker/test/SelectPickerSpec.js
```diff
@@ -400,4 +400,24 @@ describe('SelectPicker', () => {
...
+    const Wrapper = () => {
+      return (
+        <>
+          <button id="exit">exit</button>
+          <SelectPicker container={container} defaultOpen onSearch={handleSearch} data={data} />
+        </>
+      );
+    };
+    Wrapper.displayName = 'WrapperSelectPicker';
+    let { container } = render(<Wrapper />);
```
SevenOutman, line 414: Unnecessary wrapping. Simply `render` the fragment.

src/SelectPicker/test/SelectPickerSpec.js
```diff
@@ -400,4 +400,24 @@ describe('SelectPicker', () => {
...
+    const exit = container.querySelector('#exit');
```
SevenOutman, line 415: Use `getElementById` when possible.

SevenOutman (changes requested):

src/SelectPicker/test/SelectPickerSpec.js
myNameIsDu: The container doesn't have the getElementById method

src/SelectPicker/test/SelectPickerSpec.js
myNameIsDu: Yeah

src/SelectPicker/test/SelectPickerSpec.js
myNameIsDu: Yeah

src/SelectPicker/test/SelectPickerSpec.js
SevenOutman: `document` should be available in test scripts.

src/SelectPicker/test/SelectPickerSpec.js
```diff
@@ -400,4 +400,22 @@ describe('SelectPicker', () => {
...
+        <SelectPicker onClose={handleClose} defaultOpen onSearch={handleSearch} data={data} />
+      </>
+    );
+    const exit = container.querySelector('#exit');
+
+    // close select
+    fireEvent.mouseDown(exit, { bubbles: true });
+    await waitFor(() => assert.isTrue(handleClose.calledOnce));
+
+    assert.isTrue(handleSearch.calledOnce);
```
SevenOutman, line 418: Duplicated assertions. You could wrap multiple assertions within `waitFor`.

SevenOutman (changes requested):

src/SelectPicker/test/SelectPickerSpec.js
myNameIsDu: Great writing style

Issue #2412, opened 2022-03-22
Dropdown with popover, no hover highlight


dgastudio:
### What version of rsuite are you using?

5

### What version of React are you using?

17

### What browser are you using?

chrome

### Describe the Bug

Standard dropdown
![CleanShot 2022-03-22 at 23 56 54@2x](https://user-images.githubusercontent.com/161524/159574524-8e9df289-b606-4b2a-b9db-a6955d241368.png)

Popover dropdown
![CleanShot 2022-03-22 at 23 57 21@2x](https://user-images.githubusercontent.com/161524/159574581-fde153b7-0958-4444-b765-7c514104fbca.png)

there are no hover effect applied on menu items (except nested one)

### To Reproduce

https://next.rsuitejs.com/components/dropdown/
Dropdown user with popover

SevenOutman:
Closed via #2415

PR #2415, opened 2022-03-24
Fix menu item hover highlight

hassanzohdy:
Fixing Issue #2412

### DIFF
--- a/src/Dropdown/test/DropdownMenuSpec.js
+++ b/src/Dropdown/test/DropdownMenuSpec.js
@@ -234,6 +234,25 @@ describe('<Dropdown.Menu>', () => {
     assert.isNotNull(instance.querySelector('.rs-dropdown-item-submenu'));
   });
 
+  it('Should highlight menu item when hover', () => {
+    const instance = getDOMNode(
+      <DropdownMenu>
+        <DropdownItem>1</DropdownItem>
+        <DropdownItem>2</DropdownItem>
+        <DropdownItem>3</DropdownItem>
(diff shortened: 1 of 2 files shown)


src/Dropdown/test/DropdownMenuSpec.js
```diff
@@ -234,6 +234,25 @@ describe('<Dropdown.Menu>', () => {
...
+        <DropdownItem data-testid="item-3">3</DropdownItem>
+        <DropdownItem data-testid="item-4">4</DropdownItem>
+      </DropdownMenu>
+    );
+
+    const menuItem = getByTestId('item-1');
+
+    act(() => {
+      Simulate.mouseOver(menuItem);
+    });
```
SevenOutman, line 251: Please use `@testing-library/user-event` as long as it's possible. You may use its `.hover()` method.

src/Dropdown/test/DropdownMenuSpec.js
```diff
@@ -234,6 +234,25 @@ describe('<Dropdown.Menu>', () => {
...
+
+    expect(menuItem).to.have.attr('class', 'rs-dropdown-item rs-dropdown-item-focus');
```
SevenOutman, line 253: You may use `.to.have.class()` assertion so that you don't have to write down the whole class name string.

SevenOutman (changes requested):

SevenOutman (approved):
LGTM :) Thanks!

Issue #2392, opened 2022-03-09
DatePicker - `oneTap` does not work on time input


KhanShaheb34:
### What version of rsuite are you using?

5.0.0

### What version of React are you using?

17.0.2

### What version of TypeScript are you using (if any)?

4.3.5

### What browser are you using?

Chrome

### Describe the Bug

I am using a `Form.Control` component and passed `DatePicker` as accepter with `oneTap`. The `oneTap` functionality works perfectly I want to pick a date. But if I set the format to `hh-mm` or I want to pick time, it does not work. The value is not set until `Ok` button is pressed. And if I enable `oneTap`, even the `Ok` button is not visible anymore.

### Expected Behavior

`oneTap` should work on time pickers. The value should be set whenever a time is clicked.

### To Reproduce

This snippet should reproduce the bug
```js
<Form.Control
    accepter={DatePicker}
    format="hh:mm aa"
    oneTap
/>
```

PR #2433, opened 2022-04-02
feat(DatePicker): support oneTap when only time is selected

simonguo:
Support `oneTap` in time selection mode on DatePicker.  https://github.com/rsuite/rsuite/issues/2392

```tsx
<DatePicker oneTap format="hh:mm aa"  />
```


SevenOutman (commented):
The menu is not closed after selecting a time

simonguo:
> The menu is not closed after selecting a time

Because the selection time is often clicked 3 times, including hours, minutes, seconds. If the menu is closed with every click, the experience will be poor.

SevenOutman:
> Because the selection time is often clicked 3 times

That's not what "one tap" means. Maybe one tapping is not suitable for time selection at all.

Issue #2425, opened 2022-03-30
RangeSlider picks wrong progress bar direction


raibikys:
### What version of rsuite are you using?

5.6.6

### What version of React are you using?

17.0.2

### What version of TypeScript are you using (if any)?

4.3.2

### What browser are you using?

Chrome

### Describe the Bug

Please check your documentation and you will see the issue.  Looks like this:

![image](https://user-images.githubusercontent.com/70314658/160778946-77552a66-cc41-4ab1-8829-c1fde8c2ad85.png)

It happens when you select values and then randomly select 100% value. Progress bar goes opposite direction.

### Expected Behavior

Progress bar should go down in the bound of slider.

raibikys:
@SevenOutman what are the chances of it being fixed in the near future?

SevenOutman:
> @SevenOutman what are the chances of it being fixed in the near future?

It won't be too late as we are experiencing this issue in our own apps too. Contributions are also welcome and appreciated.

PR #2435, opened 2022-04-02
fix(RangeSlider): fix RangeSlider progress bar misalignment

simonguo:
fix https://github.com/rsuite/rsuite/issues/2425


SevenOutman (changes requested):
Please add a test case/edit an existing test case to ensure that the issue is fixed.

SevenOutman (approved):

PR #2442, opened 2022-04-11
feat(Dropdown): add `open` and `defaultOpen` props

SevenOutman:
Adding controlled open state for `<Dropdown>`


simonguo (changes requested):
`open` is not controlled when Dropdown is in Navbar.
image

SevenOutman:
> `open` is not controlled when Dropdown is in Navbar. 

Dropdown behaviors within Nav should be implemented separately IMO. See discussion #2456 for detailed explanation.

simonguo (approved):

PR #2461, opened 2022-04-21
feat: formControl add valuePropName support

myNameIsDu:
formControl add valuePropName support

### DIFF
--- a/src/FormControl/FormControl.tsx
+++ b/src/FormControl/FormControl.tsx
@@ -52,7 +52,7 @@ export interface FormControlProps<P = any, ValueType = any>
 
   /** Remove field value and error message when component is unmounted  */
   shouldResetWithUnmount?: boolean;
-  
+
   /** Property name of the value of the child node  */
   valuePropName?: string;
 }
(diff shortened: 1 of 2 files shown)


simonguo:
Hi @myNameIsDu 
I'm not quite sure, what problem is it trying to solve?

myNameIsDu:
> Hi @myNameIsDu I'm not quite sure, what problem is it trying to solve?

Here's an [example](https://codesandbox.io/s/unruffled-frost-vzyiit?file=/src/index.js)

The value of the Toggle component is actually checked,If I use it in a form, I have to wrap a layer, Others, such as Checkbox Component

simonguo:
@myNameIsDu 
In use, I think the encapsulation of the Toggle component is easier to understand than the `valuePropName` prop.

myNameIsDu:
> @myNameIsDu In use, I think the encapsulation of the Toggle component is easier to understand than the `valuePropName` prop.

Yeah,I think this is a very friendly prop for me,It can help me write lesser template code, And i think if one person knows how to wrap Toggle component, He must understand what this prop is doing

SevenOutman:
I don't think a custom value prop is key to this `<Toggle>` case. We don't take `checked` from `<Toggle>` because it receives value via this "customized prop", but because *it's a checkbox*.

myNameIsDu:
> I don't think a custom value prop is key to this `<Toggle>` case. We don't take `checked` from `<Toggle>` because it receives value via this "customized prop", but because _it's a checkbox_. The core of resolving this issue is to tell `<FormControl>` whether its accepter is a radio/checkbox or a normal textbox.

Yeah, if we can handle it internally that's the perfect situation

PR #2480, opened 2022-05-03
feat(DateRangePicker): expose calendars renderTitle

Roberto14:
Exposing the already existent renderTitle for calendars in DateRangePicker to allow custom titles.

### DIFF
--- a/src/DateRangePicker/test/DateRangePickerSpec.js
+++ b/src/DateRangePicker/test/DateRangePickerSpec.js
@@ -640,7 +640,7 @@ describe('DateRangePicker', () => {
 
     const [firstTitle, secondTitle] = getAllByTestId('calendar-title');
 
-    expect(firstTitle.innerText).to.equal('1 - 2022');
-    expect(secondTitle.innerText).to.equal('2 - 2022');
+    expect(firstTitle).to.have.text('1 - 2022');
+    expect(secondTitle).to.have.text('2 - 2022');
   });
 });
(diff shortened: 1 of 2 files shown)


src/DateRangePicker/test/DateRangePickerSpec.js
```diff
@@ -624,4 +624,23 @@ describe('DateRangePicker', () => {
...
+            {date.getMonth()} - {date.getFullYear()}
+          </span>
+        )}
+      />
+    );
+
+    const [firstTitle, secondTitle] = getAllByTestId('calendar-title');
+
+    expect(firstTitle.innerText).to.equal('1 - 2022');
+    expect(secondTitle.innerText).to.equal('2 - 2022');
```
SevenOutman, line 644: Use `.textContent` instead of `.innerText` to retrieve don content. Or use `.to.have.text(value)` assertion directly.

SevenOutman (changes requested):
Hi @Roberto14, thanks for this edit with standard tests! Here's a little bit change I'd like you to make so that I can approve your changes.

src/DateRangePicker/test/DateRangePickerSpec.js
Roberto14: Hi @SevenOutman, just pushed the change requested. Thanks for the review!

SevenOutman (approved):

PR #2495, opened 2022-05-23
SidenavToggle icon doesn't match its expand status

vingeraycn:
When the `Sidenav` is expanded, the icon should be the left arrow, meaning that clicking on it will close the Sidenav, whereas in reality the icon is the right arrow in the expanded state, which makes it difficult for the user to understand the meaning of the icon in this scenario.

Here is `SidenavToggle` usage code.
image

The figure below should show the **left** arrow (in the expanded state)
image

The figure below should show the **right** arrow (in the unexpanded state)
image

There are sample: https://codesandbox.io/s/vibrant-snyder-kx3fnp?file=/src/index.tsx


SevenOutman (approved):
Thanks!

PR #2500, opened 2022-05-24
fix(CheckTreePicker): fix duplicated key when data changed (#2486)

### LINKED ISSUE #2486: CheckTreePicker does not update correctly

5.11

16.8 - 18

_No response_

Chrome, Edge

## Issue
It seems that the CheckTreePicker(CTP for short) component does not update its internal state properly when the data attribute is changed.
## My Case 
I have a custom filter for a tree, based on an additional attribute, so it makes sense to have a separate selector instead of custom filtering everything on the CTP. Before I filter everything is working correctly, but when I reassign the data attribute and try to pick an item I am getting duplicated key error and in the selected list I can see two same items instead of one(ex. two Node 7 items, even tho there is only one in the tree).

### Expected Behavior

To not have a duplicated key issue when changing the data attribute

### To Reproduce

# Steps to reproduce: 
- render CTP with initial data value
- reassign data value with filtered/rearranged value
- select an item -> leads to duplicated key error

Sandbox link - https://codesandbox.io/s/cranky-davinci-11s578?file=/src/App.tsx

superman66:
When node in data changes position, it's equivalent to changing the data. And selected this node again. Will getting duplicated key error in selected list.
image

Fix #2486

### DIFF
--- a/src/CheckTreePicker/test/CheckTreePickerSpec.js
+++ b/src/CheckTreePicker/test/CheckTreePickerSpec.js
@@ -5,6 +5,7 @@ import { getDOMNode, getInstance } from '@test/testUtils';
 import CheckTreePicker from '../CheckTreePicker';
 import { KEY_VALUES } from '../../utils';
 import { assert } from 'chai';
+import { originMockData, changedMockData } from './mocks';
 
 const itemFocusClassName = '.rs-check-tree-node-focus';
 const itemExpandedClassName = '.rs-check-tree-node-expanded';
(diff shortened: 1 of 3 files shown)


src/CheckTreePicker/test/CheckTreePickerSpec.js
```diff
@@ -611,4 +612,40 @@ describe('CheckTreePicker', () => {
...
+    const mockRenderValue = (values, checkedItems, selectedElement) => {
+      checkItems = checkedItems;
+      return selectedElement;
+    };
+    const ref = React.createRef();
+    render(<TestApp ref={ref} renderValue={mockRenderValue} />);
+
+    ReactTestUtils.act(() => {
+      ref.current.setTreeData(changedMockData);
+    });
```
SevenOutman, line 641: Use `rerender` function to re-render the component with new props.

```jsx
const { rerender } = render(<Component prop={value} />);

rerender(<Component prop={newValue} />);
```

SevenOutman, line 641: Use `rerender` function to re-render the component with new props.

```jsx
const { rerender } = render(<Component prop={value} />);

rerender(<Component prop={newValue} />);
```

src/CheckTreePicker/test/CheckTreePickerSpec.js
```diff
@@ -611,4 +612,36 @@ describe('CheckTreePicker', () => {
...
+  it('Should not has duplicated key when data changed', () => {
+    const TestApp = React.forwardRef((props, ref) => {
+      const pickerRef = React.useRef();
+      React.useImperativeHandle(ref, () => {
+        return {
+          picker: pickerRef.current
+        };
+      });
+      return <CheckTreePicker ref={pickerRef} {...props} open />;
+    });
```
SevenOutman, line 625: Redundant wrapping. Render `<CheckTreePicker>` directly

SevenOutman (changes requested):

SevenOutman (approved):

PR #2518, opened 2022-06-01
feat(toaster): add support for useToaster

simonguo:
### About `useToaster`

The implementation principle of `toaster` is to render `Message` and `Notification` to a separate DOM tree through `ReactDOM.render`.
Doing so will cause a problem, components inside `Message` and `Notification` will not be able to obtain the context value provided by `CustomProvider`, including the implementation of localization functions will become very complicated.

To solve this problem, use `ReactDOM.createPortal` instead of `ReactDOM.render` to share context. So the `useToaster` hook was added to replace toaster.

----

Before
```tsx
import { toaster } from 'rsuite';

toaster.push(<Notification>message</Notification>, {
  placement: 'topEnd'
});
 
```

After

```tsx
import { useToaster } from 'rsuite';

const toaster = useToaster();

toaster.push(<Notification>message</Notification>, {
  placement: 'topEnd'
});

```

----

### Complete example

image

```tsx
import { useToaster, CustomProvider, Notification, Uploader } from 'rsuite';
import zhCN from 'rsuite/locales/zh_CN';

const App = () => {
  const toaster = useToaster();

  const handleClick = () => {
    toaster.push(
      <Notification header="消息">
        <Uploader action="/" />
      </Notification>
    );
  };

  return (
    <div>
      <button onClick={handleClick}>push</button>
    </div>
  );
};

render(
  <CustomProvider locale={zhCN}>
    <App />
  </CustomProvider>
);
```

### DIFF
--- a/src/toaster/useToaster.ts
+++ b/src/toaster/useToaster.ts
@@ -3,23 +3,26 @@ import toaster from './toaster';
 import { ToastContainerProps } from './ToastContainer';
 import { useCustom } from '../utils';
 
+/**
+ * Hook to use the toaster
+ * @returns toaster { push, remove, clear }
+ */
 const useToaster = () => {
   const { toasters } = useCustom();
 
   return {
-    push: (message: React.ReactNode, options: ToastContainerProps = {}) => {
-      if (toasters) {
-        return toasters?.current.get(options.placement || 'topCenter')?.push(message);
-      }
-      return toaster.push(message, options);
+    push: (message: React.ReactNode, options?: ToastContainerProps) => {
+      const customToaster = toasters?.current?.get(options?.placement || 'topCenter');
+
+      return customToaster ? customToaster.push(message) : toaster.push(message, options);
     },
     remove: (key: string) => {
       toasters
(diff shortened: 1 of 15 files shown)


src/toaster/useToaster.ts
```diff
@@ -0,0 +1,27 @@
...
+import { ToastContainerProps } from './ToastContainer';
+import { useCustom } from '../utils';
+
+const useToaster = () => {
+  const { toasters } = useCustom();
+
+  return {
+    push: (message: React.ReactNode, options: ToastContainerProps = {}) => {
+      if (toasters) {
+        return toasters?.current.get(options.placement || 'topCenter')?.push(message);
```
MarvelSQ, line 12: 第一个 `?` 是多余的

MarvelSQ: 这里的api是不是应该改成
```suggestion
        toasters.push(message, options);
```
现在的api可能出现拿不到 `container` 导致消息不展示

SevenOutman (approved):

MarvelSQ (approved):

PR #2523, opened 2022-06-02
feat(Nav.Menu): add `openDirection` prop

### LINKED ISSUE #2267: Allow Specifying Direction of Submenu Expansion

The <Nav.Dropdown.Menu> component does not have a property for specifying the direction of expansion and always defaults to "right" expansion. Without being able to specify expansion direction the submenu component cannot be used with a top-right-hand hamburger menu. (See screenshot below)

screenshot

The <Nav.Dropdown.Menu> should use the same "placement" prop as <Nav.Dropdown> so a top-right-hand hamburger menu could have a submenu that uses "leftStart" placement to look correct.

SevenOutman:
Close #2267

Add `openDirection` prop for specifying the opening direction of a submenu.

`openDirection="end"` (default), submenu opens towards the tail of reading direction (right by default, left in RTL)
image

`openDirection="start"`, submenu opens towards the head of reading direction (left by default, right in RTL)
image

### DIFF
--- a/docs/pages/components/nav/en-US/index.md
+++ b/docs/pages/components/nav/en-US/index.md
@@ -39,8 +39,6 @@ Provides a list of various forms of navigation menus, which can be landscape and
 
 <!--{include:`dropdown.md`}-->
 
-> Use the `<Dropdown>` component directly when using multi-level navigation.
-
 ### With Icon
 
 <!--{include:`icon.md`}-->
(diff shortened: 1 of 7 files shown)


simonguo (changes requested):
需要在文档补充对 `openDirection` prop 的说明。

simonguo (approved):

PR #2543, opened 2022-06-14
fix:  when data changes Cannot close

myNameIsDu:
example: https://stackblitz.com/edit/react-c42wf5?file=package.json,src%2Findex.js

When click the close icon of label1, and then click the close icon of label2, It will be open all.

I think it has always existed.Then this [pr](https://github.com/rsuite/rsuite/pull/2372) exposed it.

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -20,6 +20,7 @@ const webpackConfig = {
       '@test': path.resolve(__dirname, './test')
     }
   },
+  devtool:'inline-source-map',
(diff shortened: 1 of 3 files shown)


simonguo (changes requested):
Changed code fails prettier checks.
image

myNameIsDu:
> Changed code fails prettier checks. image

done

simonguo (approved):

PR #2547, opened 2022-06-17
fix(Table): fix words being split when wordWrap

simonguo:
before
image

after

image


SevenOutman:
@simonguo What about making it possible to control the wrapping behavior via this prop?

E.g.

```jsx 
<Table wordWrap="break-all" />
<Table wordWrap="break-word" />
<Table wordWrap /> // fallback to "break-all"
```

That way users who expect the current break-all wrapping don't receive breaking change.

simonguo:
👌， Changing the value of `wordWrap` makes it more flexible.

PR #2571, opened 2022-07-04
chore: add bump docs version script

SevenOutman:
Instead of manually bumping the rsuite dependency in docs after releasing, now use this `bump-docs` script

    npm run bump-docs

### What this script does

1. Update `docs/package.json` version to the same as root `package.json`
2. Update `docs/package.json` rsuite dependency version to the same as root `package.json`
3. Run `npm install` to install the latest rsuite
4. Double-check that installed rsuite version matches the root `package.json`
5. Commit changes made

### Example output

image


scripts/bumpDocsVersion.js
```diff
@@ -0,0 +1,70 @@
...
+  const docsPackageJson = await PackageJson.load(path.join(projectRoot, 'docs'));
+
+  const rsuiteVersion = rsuitePackageJson.content.version;
+
+  console.info('Local rsuite version is: ' + rsuiteVersion);
+
+  // Two steps to bump rsuite in docs
+  // 1. update docs version to the same as the latest rsuite version
+  // 2. install latest rsuite version in docs
+  console.log();
```
simonguo, line 21: Why is there an empty console.log output

SevenOutman: To add an empty line before the message for better output formatting.

simonguo (approved):

PR #2592, opened 2022-07-15
fix(SelectPicker): add caretAs prop to SelectPickerProps interface

diegopedro94:
The `caretAs` prop is missing on the `SelectPickerProps`, because of this when you use the `SelectPicker` component in a TS project, it gives an error that `caretAs` is not a valid prop.

The docs specify the `SelectPicker` component admits the prop in question.

https://rsuitejs.com/components/select-picker/#props

### DIFF
--- a/src/CheckPicker/CheckPicker.tsx
+++ b/src/CheckPicker/CheckPicker.tsx
@@ -49,7 +49,7 @@ export type ValueType = (number | string)[];
 export interface CheckPickerProps<T>
   extends FormControlPickerProps<T[], PickerLocale, ItemDataType<T>>,
     MultipleSelectProps<T>,
-    Pick<PickerToggleProps, 'label'> {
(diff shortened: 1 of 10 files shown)


diegopedro94:
@SevenOutman @simonguo If you think this is worth reviewing, it would be good to handle this to avoid using `ts-ignore`. If needed I can create a github issue to link to this PR.

simonguo (changes requested):
@diegopedro94   Thanks for your fix.  You also need to add a test for the fix.
https://github.com/rsuite/rsuite/blob/main/src/SelectPicker/test/SelectPicker.test.tsx

Also, other Picker (CheckPicker, TreePicker ...) components have the same problem except SelectPicker. Of course they can be re-opened a PR to fix.

simonguo (approved):

PR #2614, opened 2022-07-26
fix(DateRangePicker): fix predefined `ranges` cannot close picker

simonguo:
before

https://user-images.githubusercontent.com/1203827/180970534-18cdc337-e24f-40fd-837c-8a6bb4e97515.mp4

after

https://user-images.githubusercontent.com/1203827/180970878-cce711b8-b403-498f-9abb-45e190b6b312.mp4


SevenOutman:
I don't think a selecting a predefined range should close DateRangePicker immediately as selecting two dates separately doesn't close the picker either.

simonguo:
The predefined range is actually a shortcut option. From the user's point of view, I know what "today" and "yesterday" are, and there is no need for a second confirmation. It should be possible to get users to their destination with as few clicks as possible. That's what we did by default in v4, but we changed it in v5 and it's actually a missing bug.

SevenOutman:
> It should be possible to get users to their destination with as few clicks as possible.

This behavior actually adds an extra click for users who want to check the date range after clicking a range shortcut, especially when the text of the shortcut were ambiguous.

I'd suggest adding an optional property - either on DateRangePicker component or the range config object - to avoid changing the current behavior while allow developers to choose the closing behavior at the same time.

simonguo:
https://rsuitejs.com/components/date-range-picker/#code-ts-range-code
You can use `closeOverlay` to configure whether the predefined range closes the overlay after clicking.
It has been supported since rsuite v4.  The default value is `true`. The influential bootstrap daterangepicker also uses this UI interaction.

PR #2661, opened 2022-08-16
refactor(virtualized): use react-window instead of react-virtualized

simonguo:


abdullaev388:
will you also add possibility to add infinite loading. there is another package for it:
https://www.npmjs.com/package/react-window-infinite-loader

simonguo:
> will you also add possibility to add infinite loading? there is another package for it: https://www.npmjs.com/package/react-window-infinite-loader

react-window-infinite-loader will not be added, an example of infinite loading will be provided.

PR #2672, opened 2022-08-25
fix(InputNumber): fix pattern props not working

maxpou:
## steps to reproduce

1. go to https://rsuitejs.com/components/input-number/
2. replace Input component with:
```jsx
 <InputNumber pattern="[0-9]"/>
```
3. type `12.3` (note: there's a dot)

## Actual

InputNumber has value `12.3`

## Expected

InputNumber value does not contains a dot.
Numbers with decimals shouldn't be allowed, as defined in the pattern.

### DIFF
--- a/src/InputNumber/InputNumber.tsx
+++ b/src/InputNumber/InputNumber.tsx
@@ -275,7 +275,7 @@ const InputNumber = React.forwardRef((props: InputNumberProps, ref) => {
     <Input
       {...(htmlInputProps as Omit<React.InputHTMLAttributes<HTMLInputElement>, 'size'>)}
       type="number"
-      autoComplete="off"
+      autoComplete={htmlInputProps?.pattern ? 'hidden' : 'off'}
       step={step}
       inputRef={inputRef}
       onChange={handleChange}


src/InputNumber/InputNumber.tsx
```diff
@@ -275,7 +275,7 @@ const InputNumber = React.forwardRef((props: InputNumberProps, ref) => {
     <Input
       {...(htmlInputProps as Omit<React.InputHTMLAttributes<HTMLInputElement>, 'size'>)}
       type="number"
-      autoComplete="off"
+      autoComplete={htmlInputProps?.pattern ? "hidden" : "off"}
```
simonguo, line 278: The code here doesn't solve entering a decimal point.  Also you need to format the code.  

You can run the command to check the code.
```
npm run lint   // or npm run test
```

If you just want the InputNumber to not be able to enter decimals, you can format the value in onChange .

```tsx
const App = () => {
  const [value, setValue] = React.useState(1);

  return (
    <InputNumber
      value={value}
      onChange={value => {
        setValue(parseInt(value));
      }}
    />
  );
};
```

simonguo (changes requested):

src/InputNumber/InputNumber.tsx
maxpou: hi Simon,  I just linted the code and pushed.

maxpou: > The code here doesn't solve entering a decimal point.

My bad you're right. `pattern` attribute does not seem to work on input type=number. The code I wrote was probably browser-specific.

Will close the PR then.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#pattern

PR #2715, opened 2022-09-13
Add test case and documentation for `ref` attribute of `Button`

SevenOutman:
Add "Accessing DOM" section for `Button` documentation. https://rsuite-nextjs-git-test-button-ref-rsuite.vercel.app/components/button/#accessing-dom

image

### DIFF
--- a/docs/pages/components/button/en-US/index.md
+++ b/docs/pages/components/button/en-US/index.md
@@ -91,6 +91,18 @@ The buttons are laid out horizontally in the button set and are equally wide.
 
 <!--{include:`justified.md`}-->
 
+## Accessing DOM
+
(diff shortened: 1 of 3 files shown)


simonguo (approved):
It would be better if the introduction to accessing the DOM could be placed in a separate guide. https://rsuitejs.com/guide/refs-and-dom/ 

It's kind of like https://reactjs.org/docs/refs-and-the-dom.html

SevenOutman:
> It would be better if the introduction to accessing the DOM could be placed in a separate guide. 
>

Different components have different `ref` behaviors, I think it's necessary to document it for each component separately.

simonguo:
👌

PR #2718, opened 2022-09-14
fix(Stack): fix spacing compat for chrome under 84

MarvelSQ:
## ISSUES

`spacing` under **chrome 71**
截屏2022-09-14 上午9 58 22

flexbox-gap compat
- Chrome 84+
- Safari 14.1+
> compatibility table https://caniuse.com/flexbox-gap

## CHANGES
add `isSupportFlexGap` for Stack to check browser support

### DIFF
--- a/src/utils/BrowserDetection.ts
+++ b/src/utils/BrowserDetection.ts
@@ -17,7 +17,7 @@ export const isEdge = () => canUseDOM && !isIE() && 'styleMedia' in window;
 
 export const getChromeVersion = () => {
   if (canUseDOM) {
-    const match = window.navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)\./); 
+    const match = window.navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)\./);
     return match ? parseFloat(match[2]) : false;
   }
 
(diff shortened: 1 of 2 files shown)


src/utils/BrowserDetection.ts
```diff
@@ -14,3 +14,47 @@ export const isIE11 = () =>
 
 // Edge 20+
 export const isEdge = () => canUseDOM && !isIE() && 'styleMedia' in window;
+
+export const getChromeVersion = () => {
+  if (canUseDOM) {
+    const match = window.navigator.userAgent.match(/Chrome\/([0-9]+)\./);
+    return match ? parseFloat(match[1]) : false;
```
simonguo, line 21: Chromium is used in linux desktop systems.

```suggestion
    const match = window.navigator.userAgent.match(/Chrom(e|ium)\/([0-9]+)\./); 
    return match ? parseFloat(match[2]) : false;
```

src/utils/BrowserDetection.ts
```diff
@@ -14,3 +14,47 @@ export const isIE11 = () =>
...
+
+  return false;
+};
+
+export const getSafariVersion = () => {
+  if (canUseDOM) {
+    const match =
+      !!window.navigator.userAgent.match(/Safari\//) &&
+      window.navigator.userAgent.match(/Version\/([0-9]+)\./);
+    return match ? parseFloat(match[1]) : false;
```
simonguo, line 32: image

I recommend using `window.navigator.userAgent.match(/Version\/([\d\.]+).*Safari/) `. The version information it gets is more accurate. But I'm not sure if it has compatibility issues.

simonguo (changes requested):
Add tests as new code whenever possible。

src/utils/BrowserDetection.ts
MarvelSQ: there is a similar regexp from [detect-browser](https://github.com/DamonOehlman/detect-browser), the difference is their version support `_`. I think the current regexp works fine.
> https://github.com/DamonOehlman/detect-browser/blob/master/src/index.ts#L166

simonguo: 👌

PR #2732, opened 2022-09-19
fix(Stack): filter children that should not be rendered

### LINKED ISSUE #2731: Stack render false child

5.18

17.0.2

4.x

Chrome

### Describe the Bug

```ts
<Stack>
  normal children
  {false expression}
  normal children  
</Stack>
```
in dom, the `false expression` is rendered as a regular Item that takes the same space. with `spacing` or `divider` props being set, there is a bigger gap in between

### Expected Behavior

the `false expression` won't be rendered in dom

### To Reproduce

https://codesandbox.io/s/falling-mountain-h7xmzz?file=/src/index.js

MarvelSQ:
## CHANGES
use `React.Children.toArray` remove empty children like `null`, `undefined` & `boolean` value.

## ISSUE
close #2731

### DIFF
--- a/src/Stack/test/StackSpec.js
+++ b/src/Stack/test/StackSpec.js
@@ -58,4 +58,18 @@ describe('Stack', () => {
     assert.equal(getByTestId('test').children.length, 3);
     assert.equal(getByTestId('test').children[1].textContent, '|');
   });
+
+  it('Should not render empty child', () => {
+    const { getByTestId } = render(
+      <Stack data-testid="test">
+        {0}
+        {null}
+        {false}
+        {''}
+        {[1, 2]}
+      </Stack>
+    );
+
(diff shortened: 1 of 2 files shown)


src/Stack/test/StackSpec.js
```diff
@@ -58,4 +58,18 @@ describe('Stack', () => {
...
+      <Stack data-testid="test">
+        {0}
+        {null}
+        {false}
+        {''}
+        {[1, 2]}
+      </Stack>
+    );
+
+    expect(getByTestId('test').children.length).to.equal(4);
```
simonguo, line 73:
```suggestion
    expect(getByTestId('test').children).to.length(4);
```

simonguo (changes requested):

simonguo (approved):

Issue #2733, opened 2022-09-19
The last item of CheckTreePicker is invisible if virtualized and renderExtraFooter


begoat:
### What version of rsuite are you using?

5.18.1

### What version of React are you using?

18.2.0

### What browser are you using?

Chrome latest

### Describe the Bug

When `<CheckTreePicker />` received both `virtualized` and `renderExtraFooter` props, the last item is invisible.

<img width="1184" alt="image" src="https://user-images.githubusercontent.com/22862294/190946691-9b8edbae-208e-4e8a-a21a-26766b4cf582.png">

data we provided with `Picker` ranges from demo0 --- demo299

### Expected Behavior

All options are visible

### To Reproduce

https://codesandbox.io/s/adoring-gould-755g2n?file=/src/App.js:0-511

simonguo:
Fixed in  #2745

Fixed example: https://codesandbox.io/s/strange-davinci-6e7ssd
![image](https://user-images.githubusercontent.com/1203827/191488052-e3371951-b06b-4ec5-8150-4fc496329763.png)

PR #2745, opened 2022-09-21
fix(TreePicker,CheckTreePicker): fix `renderExtraFooter` causing the …

simonguo:
…last option to be invisible

fix: https://github.com/rsuite/rsuite/issues/2733

https://github.com/rsuite/rsuite/pull/1476/files#diff-28c7e8e25892c192e8b1d5e190e7aed969d6f91a5e50a7505e9a4fe946201fe4R153-R154
https://github.com/rsuite/rsuite/pull/1476/files#diff-28c7e8e25892c192e8b1d5e190e7aed969d6f91a5e50a7505e9a4fe946201fe4R153-R154

@SevenOutman `display: flex` is set for the menu in both TreePicker and CheckTreePicker styles. This is the main reason why `renderExtraFooter` is not visible. When I remove `display: flex` , the style of the component doesn't change, so I don't know what is the purpose of setting `display: flex` on the menu in TreePicker and CheckTreePicker? Are there other risks if I remove it?

### DIFF
--- a/docs/pages/components/check-picker/fragments/extra-footer.md
+++ b/docs/pages/components/check-picker/fragments/extra-footer.md
@@ -44,7 +44,6 @@ const App = () => {
         renderExtraFooter={() => (
           <div style={footerStyles}>
             <Checkbox
-              inline
               indeterminate={value.length > 0 && value.length < allValue.length}
               checked={value.length === allValue.length}
               onChange={handleCheckAll}
--- a/docs/pages/components/check-tree-picker/en-US/index.md
+++ b/docs/pages/components/check-tree-picker/en-US/index.md
@@ -42,6 +42,10 @@ The cascade attribute can set whether or not CheckTreePicker can consider the ca
 
 <!--{include:`async.md`}-->
 
+### Extra footer
+
+<!--{include:`extra-footer.md`}-->
+
 ## Accessibility
 
 Learn more in [Accessibility](/guide/accessibility).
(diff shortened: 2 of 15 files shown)


SevenOutman:
> @SevenOutman `display: flex` is set for the menu in both TreePicker and CheckTreePicker styles. This is the main reason why `renderExtraFooter` is not visible. When I remove `display: flex` , the style of the component doesn't change, so I don't know what is the purpose of setting `display: flex` on the menu in TreePicker and CheckTreePicker? Are there other risks if I remove it?

The `display: flex` style was added before #1476. That PR was simply renaming the stylesheet files.

hiyangguo:
@simonguo I review the code.If you delete `display:flex` rule from `.rs-picker-menu` in `<CheckTreePicker/>`，it will change to default rule `display:block`. And it will show the same appearance ,even the rules are not same.I think your code is safe.
But I think the real reason of this bug maybe is ` defaultHeight={inline ? height : maxTreeHeight}`.
https://github.com/rsuite/rsuite/pull/2745/files#diff-e7f6213f721793e010931abb7589019403121edde27e935bf3fe07af27031037R782
https://github.com/rsuite/rsuite/pull/2745/files#diff-8b045fcb92febebcc963e692968b0df8c8c6aa3739fff8f6c70531df07bcad5bR816
So I try to rollnack all your modifyed for all `.less` files. And result is all work fine. So you mayby don't need to modify any style files.

simonguo:
@hiyangguo  Is it possible to open a separate PR to remove `display:flex` to make the code cleaner?

hiyangguo (approved):

PR #2776, opened 2022-09-29
docs(next): fix vercel/next build failure

simonguo:
- build(deps): bump next from 12.1.0 to 12.1.6
- docs: use react-select instead of react-text-mask


superman66:
The main branch has the same problem when building on vercel.

simonguo:
@superman66   
The reason for the build failure is that there are version conflicts or out-of-date versions of multiple libraries.
`react-virtualized` is still used under the main branch, so this fix does not completely solve the problem under the main branch. Either find a way to ignore the problem in vercel, or wait for the feat/react-18 branch to be merged into the main branch.

PR #2781, opened 2022-09-30
build(docs): use local rsuite in preview build

superman66:


simonguo (changes requested):
The `tsconfig.local.json` configuration is not used in the v4 branch.
https://github.com/rsuite/rsuite/blob/b90b1d222dde513fd2a9c8b795e84ccbdc88521f/docs/next.config.js#L196-L198

PR #2787, opened 2022-10-08
build(docs): bump react and react-dom to 17.0.2

SevenOutman:


SevenOutman:
Closing because Vercel build failure can be resolved by adding `--legacy-peer-deps` flag to install commands.

PR #2801, opened 2022-10-11
Content property as shorthand to children

thing18:
allow to write code like <Button content='click me'/> instead of <Button>click me</Button>

### DIFF
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -39,6 +39,9 @@ export interface ButtonProps extends WithAsProps, React.HTMLAttributes<HTMLEleme
 
   /** Defines HTML button type attribute */
   type?: 'button' | 'reset' | 'submit';
+  
+  /** Shorthand for children */
+  content?: React.ReactNode;
 }
 
 const Button: RsRefForwardingComponent<'button', ButtonProps> = React.forwardRef(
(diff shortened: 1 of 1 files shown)


SevenOutman:
Hi @thing18, thanks for this edit. However the benefit in this edit prop doesn't look so obvious to me. Could you explain how it will make things easier than children?

thing18:
allows writing lesser code, for example
<Button content='click me'/> instead of <Button>click me</Button>
in my opinion the first version is easier to write and to write, also is not a forced method, a developer can select either of versions.

thing18:
allows writing lesser code, for example
`<Button content='click me/>` instead of `<Button>click me<Button/>`
in my opinion the first version is easier to write and to write, also is not a forced method, a developer can select either of versions.

SevenOutman:
> <Button content='click me/> instead of <Button>click me<Button/>
> in my opinion the first version is easier to write and to write

I'm afraid the second way is more familiar and straightforward to most web developers as it resembles how HTML is composed.

>  also is not a forced method, a developer can select either of versions.

I don't think I see the necessity in adding this new prop that is identical to an existing prop. I would suggest you wrap your own Button (and other) component if you prefer attribute-ish style over children.

```jsx 
function MyButton({ content, children, ...props }) {
  return <Button {...props}>{content}</Button>;
}
```

PR #2802, opened 2022-10-12
fix: when shouldResetWithUnmount should remove value and error

myNameIsDu:
Look at this [example](https://stackblitz.com/edit/react-mpmto6?file=src%2FApp.jsx,package.json,src%2Findex.js), Change the `username` input, and then close the toggle.

When the children component unmounts, Its called function is an old rending frame, so if two fields have linkage logic when one field changes, and then the other field unmounts, it will override.

### DIFF
--- a/src/Form/Form.tsx
+++ b/src/Form/Form.tsx
@@ -1,6 +1,5 @@
-import React, { useMemo, useCallback, useState, useImperativeHandle, useRef } from 'react';
+import React, { useMemo, useCallback, useImperativeHandle, useRef } from 'react';
 import PropTypes from 'prop-types';
-import isUndefined from 'lodash/isUndefined';
 import omit from 'lodash/omit';
 import { Schema, SchemaModel } from 'schema-typed';
 import type { CheckResult } from 'schema-typed';
@@ -13,6 +12,7 @@ import FormHelpText from '../FormHelpText';
 import { WithAsProps, TypeAttributes, RsRefForwardingComponent } from '../@types/common';
 import { useFormClassNames } from './useFormClassNames';
 import useSchemaModel from './useSchemaModel';
+import { useControlled } from '../utils';
 
 export interface FormProps<
   T = Record<string, any>,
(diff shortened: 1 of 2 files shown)


simonguo (commented):
Use the latest code from the `main` branch to allow ci/codesandbox to build successfully.

src/FormControl/test/FormControlSpec.js
```diff
@@ -223,35 +223,30 @@ describe('FormControl', () => {
...
-    assert.deepEqual(refValue, { username: 'test', email: '' });
+    fireEvent.change(container.querySelector('#username'), { target: { value: 'username' } });
+    assert.deepEqual(refValue, { username: 'username', email: '' });
     assert.deepEqual(refError, { username: 'The length cannot exceed 2' });
-    fireEvent.click(container.querySelector('#unmount'));
-    assert.deepEqual(refValue, { email: '' });
-    assert.deepEqual(refError, {});
+    fireEvent.change(container.querySelector('#email'), { target: { value: 'email' } });
+    assert.deepEqual(refValue, { email: 'email' });
+    assert.deepEqual(refError, { email: 'The length cannot exceed 2' });
```
simonguo, line 249: The assertion in the test case has no checks to remove value and error.

src/Form/Form.tsx
```diff
@@ -145,13 +145,13 @@ const Form: FormComponent = React.forwardRef((props: FormProps, ref: React.Ref<F
   const [_formValue, setFormValue] = useState(formDefaultValue);
   const [_formError, setFormError] = useState(formError || {});
```
simonguo, line 147: Here `useState(formDefaultValue)` can be replaced with `useControlled(formValue, formDefaultValue)`. This way you can get the ref through the third value returned by `useControlled` .

Need to pay attention to whether there is an impact on other functions.

https://github.com/rsuite/rsuite/blob/553241eb70f3a70c3867489e6d08ab1f4e41f1e7/src/utils/useControlled.ts#L21-L40

simonguo (changes requested):

src/FormControl/test/FormControlSpec.js
myNameIsDu: I think this assertion checked it was `{username: 'username', email: ''}` becoming {email: 'email'}. That has checked that the username value has been removed, and the error is the same.

simonguo (approved):

src/Form/Form.tsx
```diff
@@ -358,15 +367,15 @@ const Form: FormComponent = React.forwardRef((props: FormProps, ref: React.Ref<F
 
   const handleFieldChange = useCallback(
     (name: string, value: any, event: React.SyntheticEvent) => {
-      const formValue = getFormValue();
+      const formValue = realFormValue;
```
simonguo, line 370: 在处理 onChange 的时候使用了 realFormValue 作为最新的值。但是在 `FormValueContext.Provider`  value 中的值依然使用的是 formValue。 将导致在 FormControl 中获取不到最新的值，因此会导致表单 blur 校验会失效。  https://github.com/rsuite/rsuite/issues/3390 。

PR #2889, opened 2022-11-10
docs(Table): add instructions for `wordWrap`

simonguo:
image

### DIFF
--- a/docs/pages/components/table/en-US/index.md
+++ b/docs/pages/components/table/en-US/index.md
@@ -277,7 +277,7 @@ Display the hidden text in its entirety when hovering over the cell.
 
 <!--{include:`word-wrap.md`}-->
 
-If you want the cell to wrap, you just need to set `wordWrap`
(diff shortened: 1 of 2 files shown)


docs/pages/components/table/en-US/index.md
```diff
@@ -277,7 +277,7 @@ Display the hidden text in its entirety when hovering over the cell.
 
 <!--{include:`word-wrap.md`}-->
 
-If you want the cell to wrap, you just need to set `wordWrap`
+> ⚠️ We do not recommend using it with `wordWrap` and `virtualized`, because `virtualized` only achieves the best performance with a fixed line height. You can use the `fullText` property to solve the problem of not displaying the full text.
```
SevenOutman, line 280:
```suggestion
> ⚠️ We do not recommend using `wordWrap` with `virtualized`, because `virtualized` only achieves the best performance with a fixed line height. You can use the `fullText` property to solve the problem of not displaying the full text.
```

SevenOutman (changes requested):

SevenOutman (approved):

Issue #2917, opened 2022-11-18
[Rate] component not working properly on mobile


Davidwalser:
### What version of rsuite are you using?

5.17.1

### What version of React are you using?

18.2.0

### What version of TypeScript are you using (if any)?

4.8.2

### What browser are you using?

Chrome, Firefox

### Chrome Desktop
Working with Rate on Desktop (Chrome) is working fine!

### Chrome Mobile
When clicking the rate (in this case smileys, but also tested with default stars), the rate component is not working as expected:
When you click on the rate on the UI (e.g. first) it displays the right state, but not actually firing the onChange event.
On the second click on the same rate (e.g. first), it still displays the same state on the UI but this time also firing the onChange event.

So tl:dr the rate component only updates on the second click, but somehow still displays some state on the UI.

### Firefox (Desktop & Mobile)
On firefox it seems like it is generally broken, only updating the state after clicking more than just twice. 

### Expected Behavior

Rate component should display the right state and handling the onChange at the same time, not just at the second click.
It should work the same as on Chrome desktop view.

### To Reproduce

- first click on first smiley -> component displays right state, but is **not** firing onChange (display ok, onChange not ok)
<img width="142" alt="image" src="https://user-images.githubusercontent.com/28783795/202701996-7445de42-b827-470d-825b-8e894fad4886.png">

- second click on first smiley -> component still displaying the right state and is firing onChange (display ok, onChnage ok)
<img width="142" alt="image" src="https://user-images.githubusercontent.com/28783795/202701996-7445de42-b827-470d-825b-8e894fad4886.png">

- click on second smiley -> component displaying wrong state (no smiley) and is firing onChange (set to 0) (display wrong,  onClick wrong)
<img width="133" alt="image" src="https://user-images.githubusercontent.com/28783795/202706995-157d4c67-fd83-41d1-aacd-f32fe574182b.png">

- click on second smiley -> component displays right state, but is **not** firing onChange (display ok, onChange wrong)
<img width="130" alt="image" src="https://user-images.githubusercontent.com/28783795/202707254-347241f1-2f85-4091-9e3a-be33cfdbc82e.png">

- another click on second smiley -> component displays right state and firing onChange (display ok, onChange wrong)
<img width="130" alt="image" src="https://user-images.githubusercontent.com/28783795/202707366-937118df-3e80-452a-8d50-17cec3bd3d74.png">

Davidwalser:
[...]

PR #2945, opened 2022-11-25
fix(Rating): fix issue with state snapshots and therefore provide decent support for mobile phones

LisaKirchhofer:
Corresponding issue can be found here: https://github.com/rsuite/rsuite/issues/2917

The problem was the following: 2 functions operating on the same state were wrapped within 1 parent function -> one of them setting a state, the other one reading the state. The reading function was expecting to get the new value from the state, set by the previous function. As react operates on snapshots, the reading function retrieved the old value instead of the new one. I took the opportunity and tried to refactor those parts.

### DIFF
--- a/src/Rate/Rate.tsx
+++ b/src/Rate/Rate.tsx
@@ -197,8 +197,6 @@ const Rate: RsRefForwardingComponent<'ul', RateProps> = React.forwardRef(
           setCharacterMap(nextCharacterMap);
           onChangeActive?.(transformCharacterMapToValue(nextCharacterMap), event);
         }
-
-        return nextCharacterMap;
       },
       [allowHalf, characterMap, onChangeActive]
     );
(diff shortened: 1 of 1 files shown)


LisaKirchhofer:
@simonguo or @SevenOutman can you please have a look? :)

simonguo:
hi @LisaKirchhofer  
You need to add tests for the fixed code.  The test script can be run by the following command:
```
M=Rate npm run tdd
```

LisaKirchhofer:
Hi @simonguo!
I did not change or extend any of the functionality. All of it is already covered in the existing code - I implemented against them. So the question is, which kind of Test do you expect?
The bug I fixed was a react specific one, rather than a functional one.
Thanks in advance, Lisa

simonguo:
The issues in #2917 are currently not covered by our test scripts. If this PR solves #2917, I hope there will be a corresponding test case. In the test script, the click and keyboard can be simulated, and the github action will also automatically run the test script in Chrome and Firefox browsers.

PR #2999, opened 2023-01-03
test: replace `done` callback with async assertions

SevenOutman:
Replace `done` callback style async test cases with `waitFor` util function.

### DIFF
--- a/src/Breadcrumb/test/BreadcrumbSpec.tsx
+++ b/src/Breadcrumb/test/BreadcrumbSpec.tsx
@@ -1,10 +1,11 @@
 import React from 'react';
 import ReactTestUtils from 'react-dom/test-utils';
-import { render } from '@testing-library/react';
+import { act, render } from '@testing-library/react';
 import { getDOMNode, getInstance } from '@test/testUtils';
 import { testStandardProps } from '@test/commonCases';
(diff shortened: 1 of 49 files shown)


src/Breadcrumb/test/BreadcrumbSpec.tsx
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
 import ReactTestUtils from 'react-dom/test-utils';
-import { render } from '@testing-library/react';
+import { act, render } from '@testing-library/react';
 import { getDOMNode, getInstance } from '@test/testUtils';
 import { testStandardProps } from '@test/commonCases';
 import sinon from 'sinon';
 import Breadcrumb from '../Breadcrumb';
+import Sinon from 'sinon';
```
simonguo, line 8: `sinon` is imported 2 times.

simonguo (changes requested):

src/Breadcrumb/test/BreadcrumbSpec.tsx
SevenOutman: Duplicated imports are removed.

simonguo (approved):

Issue #3008, opened 2023-01-09
[Whisper Component]  whisper的children应该在<></>包裹字符串的时候也报错，至少给个warning


medmin:
### What version of rsuite are you using?

5.x

### What version of React are you using?

18.x

### What version of TypeScript are you using (if any)?

4.9.3

### What browser are you using?

chrome

### Describe the Bug

// error
```javascript
<Whisper trigger="hover"  {...OtherProps}>
    "Hello World"
</Whisper>
```

**// no error, but no tooltip either**
**// 但应该报错，或者给个warning**
```javascript
<Whisper trigger="hover"   {...OtherProps}>
   </>"Hello World"</>
</Whisper>
```

### Expected Behavior

error or warning

simonguo:
Whisper 需要获取 children 的位置，导致 children 必须是一个 DOM 元素。 
感谢反馈，我们修复一下。

medmin:
Thanks

PR #3012, opened 2023-01-11
fix(Whisper): add type checking for child components

simonguo:
fix: https://github.com/rsuite/rsuite/issues/3008

### DIFF
--- a/src/@types/global.d.ts
+++ b/src/@types/global.d.ts
@@ -1,1 +1,7 @@
 declare const __DEV__: boolean;
+
+declare namespace Chai {
+  interface Assertion {
+    toHavaError(expectedMessage?: string): Assertion;
+  }
+}
--- a/src/@types/global.d.ts
+++ b/src/@types/global.d.ts
@@ -2,6 +2,6 @@ declare const __DEV__: boolean;
 
 declare namespace Chai {
   interface Assertion {
-    toHavaError(expectedMessage?: string): Assertion;
+    toHaveError(expectedMessage?: string): Assertion;
   }
 }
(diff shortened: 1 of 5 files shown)


src/@types/global.d.ts
```diff
@@ -1 +1,7 @@
 declare const __DEV__: boolean;
+
+declare namespace Chai {
+  interface Assertion {
+    toHavaError(expectedMessage?: string): Assertion;
```
SevenOutman, line 5: Potential typo here. `toHaveError`

src/Overlay/OverlayTrigger.tsx
```diff
@@ -505,6 +505,13 @@ const OverlayTrigger = React.forwardRef(
         </Overlay>
       );
     };
+
+    if (
+      (typeof children === 'object' && children.type === React.Fragment) ||
+      typeof children === 'string'
+    ) {
+      console.error('The OverlayTrigger component does not accept strings or Fragments as child.');
+    }
```
SevenOutman, line 514: - The message would be printed every time the component renders.
- Better prefix the message with "[rsuite]" so users can identify it

SevenOutman (changes requested):

src/Overlay/OverlayTrigger.tsx
```diff
@@ -505,6 +505,15 @@ const OverlayTrigger = React.forwardRef(
         </Overlay>
       );
     };
+
+    if (
+      (typeof children === 'object' && children.type === React.Fragment) ||
+      typeof children === 'string'
+    ) {
+      console.error(
+        '[rsuite]: The OverlayTrigger component does not accept strings or Fragments as child.'
```
SevenOutman, line 514: I would suggest remove the semicolon to be consistent with other warning messages.

```suggestion
        '[rsuite] The OverlayTrigger component does not accept strings or Fragments as child.'
```

SevenOutman (changes requested):

Issue #2976, opened 2022-12-12
Breadcrumbs are links without href prop


andrewtquick:
### What version of rsuite are you using?

5.23.2

### What version of React are you using?

18.2.0

### What browser are you using?

Brave, Chrome, Edge, Firefox

### Describe the Bug

All Breadcrumb Items are links, without a href prop being included

### Expected Behavior

Breadcrumbs should not be links unless an href prop is provided

### To Reproduce

```
import React from 'react';
import ReactDOM from 'react-dom';
import './styles.css';


import { Breadcrumb } from 'rsuite';

const App = () => (
  <Breadcrumb>
    <Breadcrumb.Item>Home</Breadcrumb.Item>
    <Breadcrumb.Item>Components</Breadcrumb.Item>
    <Breadcrumb.Item>Breadcrumb</Breadcrumb.Item>
  </Breadcrumb>
);

ReactDOM.render(<App />, document.getElementById('root'));
```

SevenOutman:
Hi @andrewtquick, you can add `active` flag to a Breadcrumb.Item to make it render a plain text instead of a link.

```jsx
<Breadcrumb>
  <Breadcrumb.Item active>Home</Breadcrumb.Item>
  <Breadcrumb.Item active>Components</Breadcrumb.Item>
  <Breadcrumb.Item active>Breadcrumb</Breadcrumb.Item>
</Breadcrumb>
```

andrewtquick:
I went with 

```javascript
<Breadcrumb>
  <Breadcrumb.Item as="span">Home</Breadcrumb.Item>
  <Breadcrumb.Item as="span">Components</Breadcrumb.Item>
  <Breadcrumb.Item as="span">Breadcrumb</Breadcrumb.Item>
</Breadcrumb>
```

Also, I found that if only one breadcrumb has an href prop and the other breadcrumb items don't, it works fine. But if you remove the href prop from all breadcrumb items, they all become links.

PR #3027, opened 2023-01-17
fix(Breadcrumb): fix breadcrumbs being rendered as `<span>` by default

simonguo:
fix: https://github.com/rsuite/rsuite/issues/2976


simonguo:
> Does this mean that `active` prop is no longer needed?

Referring to the design of other UI component libraries, after communicating with LeightonYao, it was decided to increase the contrast of the color of the active item.  https://github.com/rsuite/rsuite/pull/3028

SevenOutman (approved):

PR #3041, opened 2023-01-30
fix(Nav.Menu): open with hover trigger in Navbar

### LINKED ISSUE #2431: Dropdown's hover trigger doesn't work inside Navbar

5.7.0

17.0.2

_No response_

Edge, Chrome

### Describe the Bug

Setting `trigger` event to "hover" for `Dropdown` (or `Nav.Dropdown`, which extends the former) doesn't work inside `Navbar`. But if Dropdown is placed outside Navbar, then it works just fine.

At first, I thought it doesn't work because of Next.js. But after some more tests, this bug doesn't seem to related to Next.js at all. In the sandbox, you will see that I reproduce the bug without Next.js

### Expected Behavior

I supposed that `trigger="hover"` should work anywhere, outside and inside of Navbar (or any other component).

### To Reproduce

I've placed two identical Dropdown menus with one inside and one outside the Navbar.

If you change `Dropdown` and `Dropdown.Item` to `Nav.Dropdown` and `Nav.Dropdown.Item`, you'll see that the result is the same and that Nav.Dropdown works properly even when placed outside Navbar.

https://codesandbox.io/s/peaceful-einstein-gns64t

### LINKED ISSUE #2720: Navbar > Nav > Nav.Menu open by hovering of mouse
I have to click menu if I want to see menu's items now
Sometimes we want to show my navbar that can be opened by hovering of a mouse.

If i want to see items in menu, I should just click menu. but I wish menu has hovering feature so when it has hovering feature, it doesn't have arrow icon. when I make a mouse move to menu section, menu will be opened automatically

SevenOutman:
Fix #2431 
Fix #2720

### DIFF
--- a/src/Navbar/NavbarDropdown.tsx
+++ b/src/Navbar/NavbarDropdown.tsx
@@ -1,5 +1,6 @@
 import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
+import castArray from 'lodash/castArray';
 import omit from 'lodash/omit';
 import { mergeRefs, PLACEMENT_8, placementPolyfill, useClassNames } from '../utils';
 import { TypeAttributes, WithAsProps, RsRefForwardingComponent } from '../@types/common';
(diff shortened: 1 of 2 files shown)


src/Disclosure/Disclosure.tsx
```diff
@@ -83,6 +83,7 @@ const Disclosure: DisclosureComponent = React.memo((props: DisclosureProps) => {
 
   const onMouseEnter = useCallback(
     (event: React.MouseEvent<HTMLElement>) => {
+      console.log('mouseenter');
```
simonguo, line 86: Logs not removed

simonguo (changes requested):

src/Disclosure/Disclosure.tsx
SevenOutman: Removed now

simonguo (approved):

Issue #3036, opened 2023-01-27
Toggle doesn't take default value from Form when being used in Form.Control/Form.Group


pkocak:
### What version of rsuite are you using?

5.25.0

### What version of React are you using?

18.2.0

### What version of TypeScript are you using (if any)?

4.9.4

### What browser are you using?

Chrome Version 109.0.5414.119 (Official Build) (arm64)

### Describe the Bug

I am using a Form and passing initialValues via formValues. Everything works fine for now except for Toggle. I am passing true in initialValues but Toggle initially renders as unchecked. When i switch on and off formValue is being set correctly however i couldn't find any solution about Toggle's initial checked.

### Expected Behavior

When true is given initially for a Form.Control with Toggle accepter, Toggle should be defaultChecked.

### To Reproduce

[Check codesandbox](https://codesandbox.io/s/shy-worker-s635nl?file=/index.js)

simonguo:
Hi @pkocak 
`Form.Control` needs to take over the `value` and `onChange` props of the form component. But `Toggle` does not have a `value` prop, if Toggle is to be used in `Form.Control`, simple encapsulation is required.

```jsx
const ControlledToggle = React.forwardRef((props, ref) => {
  const { value, ...rest } = props;
  return <Toggle {...rest} ref={ref} checked={value} />;
});

...

<Form.Control accepter={ControlledToggle} />
```

> https://codesandbox.io/s/autumn-feather-j8kjgj

SevenOutman:
@simonguo I think it's good to add a `type` prop to Form.Control component, resembling `type` attribute of `<input>` element. When `type` is `"checkbox"`, Form.Control passes `checked` prop to the accepter component instead of `value`. In addition, Toggle can be regarded as `"checkbox"` by default for convenience.

pkocak:
Hey @simonguo, thanks for the fast answer, means too much.

I'm using your way and it works perfectly fine for now but i will leave this issue open because @SevenOutman 's comment really makes sense. However, if you want to close this issue it's all yours.

simonguo:
We can check the value of `accepter` in `Form.Control`, if it is `Toggle`, use `checked` as the value of the form field.

PR #3048, opened 2023-01-31
fix(Form.Control): fix `acceptor` prop can not accept `Toggle`

simonguo:
fix: https://github.com/rsuite/rsuite/issues/3036#issuecomment-1408365439

### DIFF
--- a/src/FormControl/test/FormControlSpec.tsx
+++ b/src/FormControl/test/FormControlSpec.tsx
@@ -7,6 +7,7 @@ import Form, { FormInstance } from '../../Form';
 import FormControl from '../FormControl';
 import FormGroup from '../../FormGroup';
 import Schema from '../../Schema';
+import Toggle from '../..//Toggle';
 
 describe('FormControl', () => {
   it('Should output a input', () => {
@@ -371,5 +372,19 @@ describe('FormControl', () => {
       assert.equal(handleError.callCount, 1);
       assert.equal(handleError.firstCall.firstArg.user, 'field require');
     });
+  });
+
+  it('Should be to initialize the state of Toggle through the form value', () => {
+    const formValue = {
+      toggle: true
+    };
+
+    render(
+      <Form formValue={formValue}>
+        <FormControl name="toggle" accepter={Toggle} />
+      </Form>
+    );
+
+    expect(screen.getByRole('switch')).to.be.checked;
   });
 });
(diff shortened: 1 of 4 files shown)


src/FormControl/test/FormControlSpec.tsx
```diff
@@ -7,6 +7,7 @@ import Form, { FormInstance } from '../../Form';
 import FormControl from '../FormControl';
 import FormGroup from '../../FormGroup';
 import Schema from '../../Schema';
+import Toggle from '../..//Toggle';
```
SevenOutman, line 10: Incorrect import path

SevenOutman (changes requested):

src/FormControl/test/FormControlSpec.tsx
simonguo, line 10:
```suggestion
import Toggle from '../../Toggle';
```

SevenOutman (approved):

ab-pm:
This breaks code that passes a custom value to the `<Toggle>` via the `checked` property:
```js
<Form.Control
  name="…"
  accepter={Toggle}
  checked={myValue}
/>
```
Previously, the form control would forward the `checked` value to the `Toggle`. Now, it does overwrite the `...rest` with the `...accepterProps` containing a `checked` value from the form context. One can workaround this by passing it as `value`, like
```js
<Form.Control
  name="…"
  accepter={Toggle}
  value={myValue}
/>
```
but that's weird to pass a `value` prop to affect the `checked` prop.

I would suggest you either
* change the order of spreads to `{...accepterProps} {...rest}` or
* you already take the `value` and `defaultValue` not directly from `props` but rather use the custom keys, i.e. `props[valueKey]` and `props[defaultValueKey]`.

SevenOutman:
> This breaks code that passes a custom value to the `<Toggle>` via the `checked` property:
> 
> ```js
> <Form.Control
>   name="…"
>   accepter={Toggle}
>   checked={myValue}
> />
> ```

Hi @ab-pm, as Form.Control now controls the `checked` value of Toggle, you may want to pass `myValue` to Toggle via `formValue` (which expectedly didn't work before).

ab-pm:
Yes, it's nice that this works now. But sometimes there are cases where you want to override the value taken from the `formValue` per Form.Control, which usually works by passing it directly to `<Form.Control value={myInputText} />`. This approach is no longer working for `Toggle` accepters, which is *not intuitive*.
(An alternative solution for #3036 would have been to align `<Toggle>` with the other input components by making it take a `value` instead of a `checked` prop, but I understand this would have been a larger breaking change)
Sure I can change all my code to use `value` instead of `checked`, or in some places I might be able to adjust the `formValue` objects, but I'd prefer not having to do either of those things.

SevenOutman:
> But sometimes there are cases where you want to override the value taken from the `formValue` per Form.Control, which usually works by passing it directly to `<Form.Control value={myInputText} />`

This use case makes sense to me. I'm going to fix it.

PR #3055, opened 2023-02-03
fix(form.control): shouldResetWithUnmount isn't working

myNameIsDu:
When several components unmount at the same time.

React will call them one by one, and the later ones will overwrite the earlier ones.

### DIFF
--- a/src/Form/Form.tsx
+++ b/src/Form/Form.tsx
@@ -312,6 +312,7 @@ const Form: FormComponent = React.forwardRef((props: FormProps, ref: React.Ref<F
        * so use Ref to get future error
        */
       const formError = omit(realFormErrorRef.current, [name]);
+      realFormErrorRef.current = formError;
       setFormError(formError);
       onCheck?.(formError);
     },
@@ -325,6 +326,7 @@ const Form: FormComponent = React.forwardRef((props: FormProps, ref: React.Ref<F
        * so use Ref to get future value
        */
       const formValue = omit(realFormValueRef.current, [name]);
+      realFormValueRef.current = formValue;
       setFormValue(formValue);
       onChange?.(formValue);
     },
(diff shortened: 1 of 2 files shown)


src/FormControl/test/FormControlSpec.tsx
```diff
@@ -254,8 +255,14 @@ describe('FormControl', () => {
...
-    assert.deepEqual(refValue, { username: 'username', email: '' });
-    assert.deepEqual(refError, { username: 'The length cannot exceed 2' });
+    fireEvent.change(container.querySelector('#password') as HTMLInputElement, {
+      target: { value: 'password' }
+    });
+    assert.deepEqual(refValue, { username: 'username', password: 'password', email: '' });
+    assert.deepEqual(refError, {
+      username: 'The length cannot exceed 2',
+      password: 'The length cannot exceed 2'
+    });
```
simonguo, line 265:
```js
 expect(refValue).to.deep.equal({ username: 'username', password: 'password', email: '' })
```
We uniformly adjust the test assertion to [BDD](https://www.chaijs.com/api/bdd/) style.

src/Form/Form.tsx
```diff
@@ -325,6 +326,7 @@ const Form: FormComponent = React.forwardRef((props: FormProps, ref: React.Ref<F
        * so use Ref to get future value
        */
       const formValue = omit(realFormValueRef.current, [name]);
+      realFormValueRef.current = formValue;
```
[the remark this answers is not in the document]
myNameIsDu: The reason isn't this sentence. it's necessary. 
And look at this [example](https://codesandbox.io/s/keen-jackson-7ikpkw?file=/src/index.js).
when "email" is true. "username" and "password" will unmount, and then React will call "removeField", But because the calls are continuous, so the latter will overwrite the earlier one.

src/FormControl/test/FormControlSpec.tsx
myNameIsDu: done

simonguo (approved):

Issue #3056, opened 2023-02-06
[Message] component onClose callback invokes on render


dduafa:
### What version of rsuite are you using?

5.25.0

### What version of React are you using?

18.2.0

### What version of TypeScript are you using (if any)?

4.9.3

### What browser are you using?

Chrome

### Describe the Bug

Hello,

The `onClose` callback on the `Message` component invokes on render

### Expected Behavior

The `onClose` callback should only be invoked when the close icon is clicked, not when the component renders

### To Reproduce

```
function App() {
  return (
      <Message
        closable={true}
        onClose={() => {
          console.log("Closed");
        }}
      >
        Example Message
      </Message>
  );
}
```

simonguo:
Hi @dduafa 

The reason for this problem is that Message has a default delay of 2 seconds to close when it is used in combination with toaster. You can fix it by adding the `duration` property:

```diff
function App() {
  return (
      <Message
        closable
+       duration={0}
        onClose={() => {
          console.log("Closed");
        }}
      >
        Example Message
      </Message>
  );
}
```

SevenOutman:
> The reason for this problem is that Message has a default delay of 2 seconds to close when it is used in combination with toaster.

Should the auto-closing logic be handled by toaster instead of the Message component?

dduafa:
> 

Thank you @simonguo, setting the `duration` worked. As per @SevenOutman's question, I only needed the Message component

simonguo:
This is an issue that needs to be fixed, auto-close should work on toaster push.

PR #3065, opened 2023-02-12
fix(Message,Notification): duration property migrated to `toaster.push` option

simonguo:
fix: https://github.com/rsuite/rsuite/issues/3056

### Before
```tsx
toaster.push(<Message duration={2000}>message</Message>);
```

### After

```tsx
toaster.push(<Message>message</Message>, {
  duration: 2000
});
```

### DIFF
--- a/src/toaster/ToastContext.ts
+++ b/src/toaster/ToastContext.ts
@@ -1,0 +1,11 @@
+import React from 'react';
+
+export interface ToastContextProps {
+  usedToaster?: boolean;
+}
+
+const ToastContext = React.createContext<ToastContextProps>({ usedToaster: false });
+
+ToastContext.displayName = 'ToastContext';
+
+export default ToastContext;
--- a/docs/pages/components/message/fragments/with-toaster.md
+++ b/docs/pages/components/message/fragments/with-toaster.md
@@ -9,7 +9,7 @@ const App = () => {
   const toaster = useToaster();
 
   const message = (
-    <Message showIcon type={type}>
+    <Message showIcon type={type} closable>
       {type}: The message appears on the {placement}.
     </Message>
   );
@@ -45,7 +45,7 @@ const App = () => {
           onChange={setPlacement}
           style={{ width: 200 }}
         />
-        <Button onClick={() => toaster.push(message, { placement })}>Push</Button>
+        <Button onClick={() => toaster.push(message, { placement, duration: 5000 })}>Push</Button>
         <Button onClick={() => toaster.remove()}>Remove</Button>
         <Button onClick={() => toaster.clear()}>Clear</Button>
       </ButtonToolbar>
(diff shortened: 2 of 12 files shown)


SevenOutman (commented):
Looks like `toaster.push()` simply injects the `duration` prop, and the closing-after-timeout logic is still handled by Message and Notification components.

Now that `duration` prop is meant to be deprecated, I don't think Message and Notification components should be aware of the auto-closing thing. The closing should be taken care of by `toaster`, which also handles the opening.

simonguo:
The default value of the `duration` of Message and Notification is different, and the `duration` property needs to be kept in the component.

src/toaster/ToastContext.ts
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export interface ToastContextProps {
+  usedToaster?: boolean;
+}
+
+const ToastContext = React.createContext<ToastContextProps>({ usedToaster: false });
```
SevenOutman, line 7: The default value of `ToastContext` could be `undefined`, and child component could simply check whether its within a `ToasterContext`. That way we don't have to add this `usedToaster` property which is always set `true` whenever `ToastContext.Provider` is used.

```js
const toastContext = useContext(ToastContext);
const usedToaster = Boolean(toastContext)
```

simonguo: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L328
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L377

In React, setting a defaultValue of `undefined` for Context is deprecated. The value prop of Context.Provider is also a required prop.

SevenOutman: Passing `undefined` does not mean considering `defaultValue` optional, but to consider `ToastContext` nullable.

```js
const ToastContext = React.createContext<ToastContextProps | undefined>(undefined);

// or use null
const ToastContext = React.createContext<ToastContextProps | null>(null);
```

The point is that a "usedToaster" looks so ambiguous as a property of a "ToastContext". When you _call `useContext(ToastContext)` and get `usedToaster: false`_, it sounds like "yes, but no".

simonguo: This would have to set a meaningless value for Context.Provider since its value property is required.
```tsx
<Context.Provider value={null}>
```

I think it is also possible without Context. Add a `usedToaster` prop to `<Message>` that is not exposed to users.

SevenOutman: > This would have to set a meaningless value for Context.Provider since its value property is required.
> 
> ```tsx
> <Context.Provider value={null}>
> ```

Not really. The value passed to `ToastContext.Provider` will not be `null` but info about the toast (defined as `ToastContextProps`) - currently none though.

```jsx
<ToastContext.Provider value={{ placement, duration, containerId, ...thingsLikeThat }}>
</ToastContext.Provider>
```
 
> Add a `usedToaster` prop to `<Message>` that is not exposed to users.

I don't think it's a good idea to add extra props around the closing behavior to Message, while we're about to remove the closing logic from it in the future.

---

To conclude, I believe current implementation is workable. I'm going to merge it as is.

SevenOutman (approved):

PR #3067, opened 2023-02-13
feat(Button): add startIcon and endIcon props

SevenOutman:
## Feature

Add `startIcon` and `endIcon` props to specify an icon element placed before/after text content in a Button.

image

Design material: https://www.figma.com/file/uEq9QBplcKUYZrcWWA3RK2/NXT-UI?node-id=0%3A2050&t=V95uYVLyuHWmOIYM-4

### How's it different from directly adding icon element in `children`?

You have to handle alignment if putting icon element as `children`, otherwise you get misaligned icons as shown below.

image

(This is an example from rsuitejs.com)

image

### Why `start/end` rather than `left/right`?

`start/end` indicates whether the icon is placed _before_/_after_ the text, which is not bound to be left/right if taking RTL situations into account.

### DIFF
--- a/src/Button/test/ButtonSpec.tsx
+++ b/src/Button/test/ButtonSpec.tsx
@@ -116,21 +116,13 @@ describe('Button', () => {
   });
 
   it('Should render `startIcon` before text', () => {
-    render(
-      <Button startIcon={<i>Icon</i>} data-testid="button">
-        Text
-      </Button>
-    );
+    render(<Button startIcon={<i>Icon</i>}>Text</Button>);
 
-    expect(screen.getByTestId('button')).to.have.text('IconText');
+    expect(screen.getByRole('button')).to.have.text('IconText');
   });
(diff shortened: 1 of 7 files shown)


src/Button/test/ButtonSpec.tsx
```diff
@@ -114,4 +114,23 @@ describe('Button', () => {
...
   });
+
+  it('Should render `startIcon` before text', () => {
+    render(
+      <Button startIcon={<i>Icon</i>} data-testid="button">
+        Text
+      </Button>
+    );
+
+    expect(screen.getByTestId('button')).to.have.text('IconText');
```
simonguo, line 125: The button element can be returned using `screen.getByRole('button')`.

SevenOutman: OK, tests updated.

simonguo (approved):

PR #3075, opened 2023-02-21
feat: added Hungarian locale (hu_HU)

kaloczikvn:
Added the Hungarian locale (hu_HU)


SevenOutman (approved):
Thanks for contributing this!

PR #3084, opened 2023-02-27
docs(examples): add remix.run example

MarvelSQ:
## CHANGES

add an example for [remix.run](https://remix.run/)


examples/with-remix/README.md
```diff
@@ -0,0 +1,20 @@
+# Remix example
+
+## How to use
+
+Download the example:
+
+```
+curl https://codeload.github.com/rsuite/rsuite/tar.gz/master | tar -xz --strip=2 rsuite-master/examples/with-remix
+cd with-reason-webpack
```
simonguo, line 9:
```suggestion
cd with-remix
```

examples/with-remix/app/root.tsx
```diff
@@ -0,0 +1,31 @@
...
+});
+
+export default function App() {
+  return (
+    <html lang="en">
+      <head>
+        <Meta />
+        <Links />
+      </head>
+      <body>
```
simonguo, line 19: It is best to add a `padding` to the body.
image

examples/with-remix/app/routes/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+            appearance="link"
+            as="a"
+            target="_blank"
+            href="https://remix.run/docs"
+            rel="noreferrer"
+          >
+            Remix Docs
+          </Button>
+        </List.Item>
+      </List>
```
simonguo, line 48: Developers often encounter how to support `Link` in navigation. Here I think it can be changed to a Nav example.
https://github.com/rsuite/rsuite/blob/main/examples/with-nextjs/components/navbar.js#L16

simonguo (changes requested):

examples/with-remix/app/root.tsx
MarvelSQ: body padding via shared style

simonguo (approved):

PR #3102, opened 2023-03-07
refactor(DatePicker,DateRangePicker): rename prop disabledDate -> isDateDisabled

SevenOutman:
Reduce ambiguousness

### DIFF
--- a/src/utils/dateUtils.ts
+++ b/src/utils/dateUtils.ts
@@ -69,10 +69,10 @@ function validTime(calendarProps: any, date: Date) {
       return calendarProps[key]?.(getHours(date), date) ?? true;
     }
     if (/(Minutes)/.test(key)) {
-      return calendarProps[key](getMinutes(date), date);
+      return calendarProps[key]?.(getMinutes(date), date) ?? true;
     }
     if (/(Seconds)/.test(key)) {
-      return calendarProps[key](getSeconds(date), date);
+      return calendarProps[key]?.(getSeconds(date), date) ?? true;
     }
     return false;
   });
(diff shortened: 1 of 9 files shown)


simonguo:
If you want to change the name of `disabledDate`, you need to consider the names of the names of `disabledHours`, `disabledMinutes`, and `disabledSeconds` also changed, so as to maintain consistency.

In addition, I recommend using the method name of the method of `shouldDisableDate`, refer to the [MUI API](https://mui.com/x/api/date-pickers/date-picker/#props).

SevenOutman:
> If you want to change the name of `disabledDate`, you need to consider the names of the names of `disabledHours`, `disabledMinutes`, and `disabledSeconds` also changed, so as to maintain consistency.

OK, I'll rename them too.

> In addition, I recommend using the method name of the method of `shouldDisableDate`, refer to the [MUI API](https://mui.com/x/api/date-pickers/date-picker/#props).

The "should" makes more sense to me. I'll use that.

simonguo (approved):
The documentation needs to be updated after the version is released.

src/utils/dateUtils.ts
```diff
@@ -66,20 +66,21 @@ function validTime(calendarProps: any, date: Date) {
 
   return Object.keys(calendarProps).some(key => {
     if (/(Hours)/.test(key)) {
-      return calendarProps[key](getHours(date), date);
+      return calendarProps[key]?.(getHours(date), date) ?? true;
```
simonguo, line 69: The default is `true`, which will cause the time option to always be in a warning state.
image

PR #2455, opened 2022-04-17
refactor(Modal,Drawer): deprecate `full` prop

SevenOutman:
See https://github.com/rsuite/rsuite/discussions/2454 for detailed explanation.

PR #3103, opened 2023-03-08
fix(Modal,Drawer): fix not full screen when `size=full`

simonguo:
## before

image

### DIFF
--- a/docs/pages/components/drawer/fragments/size.md
+++ b/docs/pages/components/drawer/fragments/size.md
@@ -28,11 +28,11 @@ const App = () => {
     <>
       <RadioGroup inline appearance="picker" value={size} onChange={setSize}>
         <span style={styles.radioGroupLabel}>Size: </span>
-        <Radio value="full">Full page</Radio>
-        <Radio value="lg">Large</Radio>
-        <Radio value="md">Medium</Radio>
-        <Radio value="sm">Small</Radio>
         <Radio value="xs">Xsmall</Radio>
+        <Radio value="sm">Small</Radio>
+        <Radio value="md">Medium</Radio>
+        <Radio value="lg">Large</Radio>
+        <Radio value="full">Full page</Radio>
       </RadioGroup>
       <hr />
       <ButtonToolbar>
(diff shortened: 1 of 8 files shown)


SevenOutman:
Should this change be considered breaking?

simonguo:
> Should this change be considered breaking?

It's a breaking change in terms of features, but I think it's a fix for the UI design.

SevenOutman:
> It's a breaking change in terms of features, but I think it's a fix for the UI design.

This is breaking to users who do expect the 60px margin and those relying on `@drawer-full-size-margin` variable. I would suggest adding a new value to `size` prop, in order to introduce this fullscreen Modal without breaking existing ones.

simonguo:
There are 2 API designs. The first is to add a `size="xl"` instead of `size="full"`. The final size property values include: `xs , sm ,md, lg, xl , full`.

Alternatively, revert this PR #2455 to make `full` the property name. Consider adding responsive property values in the future.

SevenOutman:
I vote for the first approach, because reverting #2455 will bring back the problem discussed in #2454.

However the first approach still changes the behavior of `size="full"`. What about adding a new `size` value for this new fullscreen variant, e.g. `size="cover"`. If the naming is not as satisfying we can change it in major updates then.

SevenOutman:
> It would be beneficial to allow size to support both number and string input for greater flexibility in application scenarios.

That makes great sense!
Now that custom `size` is supported, what about keeping `size="full"` original behavior, and let user specify `size="100%"` when they decide to use a full-page modal? Just in order to avoid breaking changes ;)

simonguo:
I ultimately still hope to fix the UI presentation for size="full", as its current presentation is incorrect.


Keep the original behavior of `size="full"` by using below.
```tsx
<Modal size="calc(100% - 120px)" />  
```

SevenOutman:
> I ultimately still hope to fix the UI presentation for size="full", as its current presentation is incorrect.

Alright. Then should this PR be merged into v5 or should it wait for v6?

SevenOutman (approved):

Issue #3105, opened 2023-03-08
Autocomplete component "autocomplete" and "readOnly" props


iska9der:
### What version of rsuite are you using?

5.28.1

### What version of React are you using?

18.2.0

### What version of TypeScript are you using (if any)?

4.9.4

### What browser are you using?

Chromium

### Describe the Bug

1) readOnly prop is not working, even on your site
[rsuitejs.com/components/auto-complete](https://rsuitejs.com/components/auto-complete/#disabled-and-read-only)

autoComplete prop is not forwarded to component, but if i pass `autoComplete="off"`, it will work
without prop: 
![photo_2023-03-08_19-14-48](https://user-images.githubusercontent.com/11888809/223768141-b0386a7a-7117-40d1-9b20-8b89cf9e03af.jpg)
`autoComplete="off"`:
![photo_2023-03-08_19-14-44](https://user-images.githubusercontent.com/11888809/223768177-2f116fc9-1756-4c13-a32a-a0af78c3494a.jpg)

### Expected Behavior

1) readOnly prop must be passed to the input element so that it cannot be changed.

2) autoComplete prop needs to be in AutoCompleteProps.

### To Reproduce

https://codesandbox.io/s/boring-clarke-v6my55?file=/src/index.tsx

PR #3108, opened 2023-03-09
feat(common): add autoComplete for FormControlBaseProps

imyuanx:
Can I add `autoComplete` to FormControlBaseProps?
I have noticed that many components use FormControlBaseProps, but I think this will not affect, coz all are `Input`

Link #3105

### DIFF
--- a/src/AutoComplete/AutoComplete.tsx
+++ b/src/AutoComplete/AutoComplete.tsx
@@ -61,6 +61,9 @@ export interface AutoCompleteProps<T = ValueType>
   /** The width of the menu will automatically follow the width of the input box */
   menuAutoWidth?: boolean;
 
+  /** AutoComplete Content */
+  autoComplete?: string;
+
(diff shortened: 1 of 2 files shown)


src/@types/common.ts
```diff
@@ -147,6 +147,9 @@ export interface FormControlBaseProps<
 
   /** Make the control readonly */
   readOnly?: boolean;
+
+  /** AutoComplete Content */
+  autoComplete?: string;
```
simonguo, line 152: `autoComplete` should not be added to `FormControlBaseProps`, because the props of components such as Rate and DatePicker all extend `FormControlBaseProps`. You should fix it in `AutoCompleteProps`.

simonguo (changes requested):

src/@types/common.ts
imyuanx: Done!

PR #3151, opened 2023-04-07
feat(DateRangePicker): limit start year

### LINKED ISSUE #3145: Limit start year for DateRangePicker
`DateRangePicker` component already has a property to limit max years. It would be great to also be able to limit min years. As I understood, the `MonthDropdown` already has a possibility to limit min years, so it is only one more prop to be propagated.

image

```
<DateRangePicker
          limitEndYear={1}
          limitStartYear={-5}
          ...
        />
```

loposkin:
# CHANGES

Added `limitStartYear` prop to DateRangePicker. Usage:
```
<DateRangePicker
          limitEndYear={1}
          limitStartYear={5}
          ...
>
```

This looks like: 
limit_start_year

If unset, the start year still defaults to 1900

# ISSUE

close #3145

### DIFF
--- a/src/DateRangePicker/test/CalendarMonthDropdownSpec.tsx
+++ b/src/DateRangePicker/test/CalendarMonthDropdownSpec.tsx
@@ -6,6 +6,9 @@ import { DateUtils } from '../../utils';
 
 describe('DateRangePicker-MonthDropdown', () => {
   it('Should output year and month of previous and current year', () => {
+    const currentYear = DateUtils.getYear(new Date());
+    const previousYear = currentYear - 1;
+
     const { getAllByRole } = render(
       <CalendarContext.Provider
         value={{
@@ -14,12 +17,10 @@ describe('DateRangePicker-MonthDropdown', () => {
           isoWeek: false
         }}
       >
-        <MonthDropdown show limitStartYear={1} limitEndYear={1} />
+        <MonthDropdown show limitStartYear={previousYear} limitEndYear={currentYear} />
       </CalendarContext.Provider>
     );
 
-    const currentYear = DateUtils.getYear(new Date());
-    const previousYear = currentYear - 1;
     expect(getAllByRole('row', { hidden: true })).to.be.lengthOf(2);
     expect(getAllByRole('rowheader', { hidden: true })[0].innerText).to.be.eq(
       previousYear.toString()
--- a/src/DateRangePicker/test/CalendarMonthDropdownSpec.tsx
+++ b/src/DateRangePicker/test/CalendarMonthDropdownSpec.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import CalendarContext from '../../Calendar/CalendarContext';
+import MonthDropdown from '../../Calendar/MonthDropdown';
+import { DateUtils } from '../../utils';
+
+describe('DateRangePicker-MonthDropdown', () => {
+  it('Should output year and month of previous and current year', () => {
+    const { getAllByRole } = render(
+      <CalendarContext.Provider
+        value={{
+          date: new Date(),
+          locale: {},
+          isoWeek: false
+        }}
+      >
+        <MonthDropdown show limitStartYear={1} limitEndYear={1} />
+      </CalendarContext.Provider>
+    );
+
+    const currentYear = DateUtils.getYear(new Date());
+    const previousYear = currentYear - 1;
+    expect(getAllByRole('row', { hidden: true })).to.be.lengthOf(2);
+    expect(getAllByRole('rowheader', { hidden: true })[0].innerText).to.be.eq(
+      previousYear.toString()
+    );
+    expect(getAllByRole('rowheader', { hidden: true })[1].innerText).to.be.eq(
+      currentYear.toString()
+    );
+  });
+});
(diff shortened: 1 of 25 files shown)


loposkin:
@simonguo, do I need to do something else? Looks like the autorization is missing on some test stage.

SevenOutman (changes requested):
Hi @loposkin, thank you for contributing this edit!

As this PR adds new APIs, please add test cases to ensure the API work as expected.

loposkin:
@SevenOutman, as I only added propagation of `limitStartYear` prop to existent API of `Calendar/MonthDropdown`, it's not really clear for me, where should I add tests? It looks like there are no similar tests for `limitMaxYear` as well. Should I add the test to `Calendar/CalendarMonthDropdownItemSpec`?

src/DateRangePicker/test/CalendarMonthDropdownSpec.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+    const currentYear = DateUtils.getYear(new Date());
+    const previousYear = currentYear - 1;
+    expect(getAllByRole('row', { hidden: true })).to.be.lengthOf(2);
+    expect(getAllByRole('rowheader', { hidden: true })[0].innerText).to.be.eq(
+      previousYear.toString()
+    );
+    expect(getAllByRole('rowheader', { hidden: true })[1].innerText).to.be.eq(
+      currentYear.toString()
+    );
```
SevenOutman, line 29: Let me confirm this. Does it mean that `limitStartYear={1}` results in the earliest year be 1 year before today, while `limitEndYear={1}` results in the latest year be just current year? If so, I don't think it's consistent.

loposkin: Hi, yes, you're right and yes, it's not consistent.  The limitStartYear={1} results in the earliest year that is 1 year before current, while limitEndYear={1} results in the current year. The problem is that I only propagate the props to existent `MonthDropdown`-API. Should I change this API? I think it would mean a breaking change for people who already use MonthDropdown.

simonguo:
Hi @loposkin 

If the `limitEndYear` and `limitStartYear` props are to be exposed, I hope he can make some adjustments in the design to make it more reasonable to use.

```ts
limitStartYear?: number | ((activeDate: Date) => number);
limitEndYear?: number | ((activeDate: Date) => number);
```

where the `number` value should be a year.

```tsx
<DateRangePicker limitStartYear={1990} limitEndYear={2023} />
<DateRangePicker
  limitStartYear={(activeDate: Date) => {
    return addDays(getYear(activeDate), -5);
  }}
  limitEndYear={(activeDate: Date) => {
    return addDays(getYear(activeDate), 5);
  }}
/>;
```

Also `limitEndYear` and `limitStartYear` props should be supported on both DatePicker and Calendar.

loposkin:
@simonguo  @SevenOutman , I've changed the API according to @simonguo suggestion. I've also updated `Calendar`'s and `DatePicker`'s API accordingly. `limitStartYear` and `limitEndYear` are now consistent and the api looks like:

```
<DateRangePicker
          limitEndYear={2018}
          limitStartYear={2023}
          ...
>
```

I've updated the english documentation accordingly, unfortunatelly, I am missing required language skills to update the documentation in Chinese.

As for functional values for `limitStartYear` and `limitEndYear` props, this exceeds the original issue and I have no resources to implement it right now.

SevenOutman:
> Hi, yes, you're right and yes, it's not consistent. The limitStartYear={1} results in the earliest year that is 1 year before current, while limitEndYear={1} results in the current year. The problem is that I only propagate the props to existent `MonthDropdown`-API. Should I change this API? I think it would mean a breaking change for people who already use MonthDropdown.

@loposkin, MonthDropdown is an internal component that's not exported as a public API, so I think it's OK to either change it's `limitStartYear` behavior, or simply leave it as is and change the value that DateRangePicker passes to MonthDropdown. That way `limitStartYear` of DateRangePicker (as a public API) is consistent with `limitEndYear`, and minimum changes have to be made.

---

> If the `limitEndYear` and `limitStartYear` props are to be exposed, I hope he can make some adjustments in the design to make it more reasonable to use.
> where the `number` value should be a year.
> 
> ```tsx
> <DateRangePicker limitStartYear={1990} limitEndYear={2023} />
> ```

@simonguo, the `limitEndYear` prop already exists on DateRangePicker, so a direct change like this is breaking. I don't think it's appropriate to introduce this change in the current major release. I agree that `limitStartYear` and `limitEndYear` should have been more straightforward, we can make this change in next major release with migration guidance for users.

loposkin:
@SevenOutman, so if I change the api to be consistent to existent `limitEndYear` APIs, such that `<DateRangePicker limitStartYear={1} limitEndYear={1} />` will only show current year and `<DateRangePicker limitStartYear={2} limitEndYear={2} />` will show the years from 2022 to 2024, would it be possible to merge this in the next release? Or would you rather implement it properly in the next major release?

SevenOutman:
> @SevenOutman, so if I change the api to be consistent to existent `limitEndYear` APIs, such that `<DateRangePicker limitStartYear={1} limitEndYear={1} />` will only show current year and `<DateRangePicker limitStartYear={2} limitEndYear={2} />` will show the years from 2022 to 2024, would it be possible to merge this in the next release? Or would you rather implement it properly in the next major release?

It's going to be merged.

PR #3153, opened 2023-04-08
Update ru_RU.ts

personaljs:


iska9der:
> Hi @personaljs, thanks for contributing this edit!
> 
> As this change is around a language that I'm not familiar with, I couldn't confirm whether they should be applied. My translation software gives some hint but I'm not sure I understand it as accurately. Could you explain to me why the original translations should be updated to newer ones?

Since the topic starter takes a long time to answer, I, as a native speaker, confirm the correctness of the changes

In CIS Region we use that format:

```
formattedDayPattern: 'dd MMM yyyy'
```
---

These are previous translations, with errors:
```
notSelected: 'Није изабран',
notUploaded: 'Није отпремљено'
...
skip: 'Идти к{0}'
...
progress: 'Выгрузка'
...
complete: 'Завершенно'
```
If i trying to repeat that mistakes in eng, it will be like:

```
notSelected: 'Noт select',
notUploaded: 'No sэnd'
...
skip: 'Walk to{0}'
...
progress: 'Unloading'
...
complete: 'Finnished'
```

New translations are correct

SevenOutman:
> Since the topic starter takes a long time to answer, I, as a native speaker, confirm the correctness of the changes

@iska9der Thanks for helping reviewing this! The new translation is going to be released today ;)

personaljs:
Yeah, old translation was serbian language) Thank you, Iskander)

PR #3163, opened 2023-04-17
Daterangepicker limit start year

### LINKED ISSUE #3145: Limit start year for DateRangePicker

loposkin:
# CHANGES

Added `limitStartYear` prop to `DatePicker` and `DateRangePicker`. Usage:

Display current year only:

```
<DateRangePicker
          limitEndYear={1}
          limitStartYear={1}
          ...
>
```

Display years between previous and next year:
```
<DateRangePicker
          limitEndYear={2}
          limitStartYear={2}
          ...
>
```

This looks like: 
limit_start_year

If unset, the start year still defaults to 1900

# ISSUE

close #3145


SevenOutman (commented):
@loposkin Thank you!

LGTM. I'll push a commit to add the new prop in Chinese doc, and the PR is going to be merged.

PR #3173, opened 2023-04-27
fix(CheckPicker): avoid mutating data when groupBy is specified

SevenOutman:
Deprecate the unsafe `getDataGroupBy` util function, which mutates CheckPicker's `data' prop in-place. Replace with a safe implementation.

Update CheckPicker to make use of the safe version `getDataGroupBy`.


simonguo (changes requested):
Clicking on the group name does not work.

image


> https://rsuite-nextjs-git-fix-checkpicker-avoid-mutating-data-rsuite.vercel.app/components/check-picker/#thread-id=h3zCU

simonguo (approved):

PR #3186, opened 2023-05-09
refactor(SelectPicker,InputPicker): drop usage of UNSAFE_getDataGroupBy

SevenOutman:
Replace `UNSAFE_getDataGroupBy` with the safe implementation


simonguo (changes requested):
The button to collapse groups is not work
image

SevenOutman:
> The button to collapse groups is not work

It's fixed now

simonguo (approved):

PR #3188, opened 2023-05-09
chore: add lint cache

Simon-He95:


SevenOutman (approved):
Hi @Simon-He95, thanks for contributing to rsuite! This cache trick is helpful!

PR #3195, opened 2023-05-12
fix(DatePicker): added stack styles to datepicker styles

loposkin:
DatePicker and DateRangePicker were missing Stack styles. Therefore, the pickers were broken when using modularized import.

```
import "rsuite/DateRangePicker/styles/index.less";
```

Before:

styles-2

After:
styles-1

### DIFF
--- a/src/DatePicker/styles/index.less
+++ b/src/DatePicker/styles/index.less
@@ -2,6 +2,7 @@
 @import '../../styles/mixins/listbox';
 @import '../../Button/styles/mixin';
 @import '../../Picker/styles/mixin';
+@import '../../Stack/styles/index';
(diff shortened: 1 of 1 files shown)


loposkin:
Looks like it changes code coverage. Do you have an idea how to mitigate that?

SevenOutman (approved):
LGTM. Thanks!

SevenOutman:
> Looks like it changes code coverage. Do you have an idea how to mitigate that?

It turned out that the coverage decreased because some coverage reported failed to be uploaded in CI job due to Codecov server error. The codebase is alright. Re-running the CI jobs should resolve this.

Thanks for being careful!

PR #3199, opened 2023-05-18
docs: updated version of @faker-js/faker

alecodev:
Updated version of @faker-js/faker due to sandbox bugs and version differences.

image


simonguo (approved):
Thanks

PR #3204, opened 2023-05-19
test: [DRAFT] test component by cypress with vite

Sleaf:


simonguo:
Hi @Sleaf  switching from Karma to Cypress is a significant project. What problems will this change solve, and what benefits will it bring?

Sleaf:
- Cypress use browser environment (it supports `chrome`, `electron`, `edge`, `firefox` so far) instead of JSDom , so you can test component style or mock user behavior easily.
- Syntax of writing Cypress test case is linear and human reading friendly. 
- Cypress has a nice dev UI, and it can also run with headless mode.

It's not a well-thought-out plan, this PR is just an example for @SevenOutman to show Cypress look like now.

There are many details that may need to be discussed.

PR #3046, opened 2023-01-31
test: drop use of getDOMNode util

SevenOutman:
Drop usage of `getDOMNode()` util function which uses [the deprecated `findDOMNode` API](https://beta.reactjs.org/reference/react-dom/findDOMNode) under the hood

## Progress and plan

There're still 617 usage of `getDOMNode()` across 139 files, per search result.

image

I'm thinking about marking this PR as finished at current progress, and make the rest progress open to community contributions. It would be an opportunity for community contributors to get familiar with unit testing in rsuite codebase.

I've opened a discussion #3200 to describe what needs to be done and how should them be done, and break the whole task into small pieces so contributors can adopt progressively.

PR #3205, opened 2023-05-19
Feat/remove find dom node

restartgr:
Drop usage of getDOMNode() util function which uses [the deprecated findDOMNode API](https://beta.reactjs.org/reference/react-dom/findDOMNode) under the hood

I received the task from discussion https://github.com/rsuite/rsuite/discussions/3200

I have completed these components (DateRangePicker,Divider,Drawer,Dropdown,FlexboxGrid,Footer,Form)

Thank you for reviewing my work

### DIFF
--- a/src/Drawer/Drawer.tsx
+++ b/src/Drawer/Drawer.tsx
@@ -58,7 +58,6 @@ const Drawer: DrawerComponent = React.forwardRef((props: DrawerProps, ref) => {
   } = props;
   const { merge, prefix } = useClassNames(classPrefix);
   const classes = merge(className, prefix(placement));
-
   const animationProps = {
     placement
   };
--- a/src/Drawer/Drawer.tsx
+++ b/src/Drawer/Drawer.tsx
@@ -58,6 +58,7 @@ const Drawer: DrawerComponent = React.forwardRef((props: DrawerProps, ref) => {
   } = props;
   const { merge, prefix } = useClassNames(classPrefix);
   const classes = merge(className, prefix(placement));
+
   const animationProps = {
     placement
   };
--- a/src/Drawer/test/DrawerActionsSpec.tsx
+++ b/src/Drawer/test/DrawerActionsSpec.tsx
@@ -10,5 +10,6 @@ describe('Drawer.Actions', () => {
     const title = 'Test';
     const { container } = render(<Drawer.Actions>{title}</Drawer.Actions>);
     expect(container.firstChild).to.have.text(title);
+    expect(container.firstChild).to.have.class('rs-drawer-actions');
   });
 });
--- a/src/Drawer/test/DrawerActionsSpec.tsx
+++ b/src/Drawer/test/DrawerActionsSpec.tsx
@@ -1,29 +1,14 @@
 import React from 'react';
-import { getDOMNode } from '@test/testUtils';
-
+import { render } from '@testing-library/react';
 import Drawer from '../index';
+import { testStandardProps } from '@test/commonCases';
 
 describe('Drawer.Actions', () => {
+  testStandardProps(<Drawer.Actions />);
+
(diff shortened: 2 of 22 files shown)


src/Drawer/test/DrawerHeaderSpec.tsx
```diff
@@ -1,44 +1,33 @@
...
+    expect(container.firstChild).to.have.class('rs-drawer-header');
+    expect(container.firstChild).to.have.text(title);
   });
 
   it('Should hide close button', () => {
     const title = 'Test';
-    const instance = getDOMNode(<Drawer.Header closeButton={false}>{title}</Drawer.Header>);
-    assert.isNull(instance.querySelector('button'));
+    const { container } = render(<Drawer.Header closeButton={false}>{title}</Drawer.Header>);
+    expect(container.firstChild).to.not.have.class('button');
```
SevenOutman, line 21: The old assertion was checking _a `<button>` does not exists inside of Drawer.Header_, but the new assertion is checking _Drawer.Header does not have `.button` class name_.

src/Drawer/Drawer.tsx
```diff
@@ -58,7 +58,6 @@ const Drawer: DrawerComponent = React.forwardRef((props: DrawerProps, ref) => {
   } = props;
   const { merge, prefix } = useClassNames(classPrefix);
   const classes = merge(className, prefix(placement));
-
```
SevenOutman, line 61: Please remove this change from this PR as it's irrelevant.

src/Dropdown/test/DropdownMenuSpec.tsx
```diff
@@ -94,10 +93,10 @@ describe('<Dropdown.Menu>', () => {
...
           </DropdownMenu>
         );
-
-        fireEvent.keyDown(menubar, { key: 'ArrowDown' });
-
-        expect(menubar.getAttribute('aria-activedescendant')).to.equal('second-item');
+        if (menubar) {
+          fireEvent.keyDown(menubar, { key: 'ArrowDown' });
+          expect(menubar).to.have.attr('aria-activedescendant', 'second-item');
+        }
```
SevenOutman, line 99: The `if` would cause the test case to incorrectly pass when `menubar` is `null` - which is expected. To avoid type error here, you could explicitly declare `renderMenubar` to return an non-null HTMLElement

test/commonCases.js
SevenOutman, line 1: I noticed that you adopted changes in `test/commonCases.js` from #3046. Please also include changes in [`NavItemSpec.tsx`](https://github.com/rsuite/rsuite/pull/3046/files#diff-eef182f79781f91023790ded377d544bce788255118e2dc2a88045e737577e3a) and [`SidenavToggleSpec.tsx`](https://github.com/rsuite/rsuite/pull/3046/files#diff-1eaa9e7cf82bbaf6eed0ae3121a0c7a4dabf7e38695e3a2898394fbcaaee3689) so that the tests won't be broken.

src/Drawer/test/DrawerActionsSpec.tsx
```diff
@@ -1,29 +1,14 @@
...
-    const fontSize = '12px';
-    const instance = getDOMNode(<Drawer.Actions style={{ fontSize }} />);
-    assert.equal(instance.style.fontSize, fontSize);
-  });
-
-  it('Should have a custom className prefix', () => {
-    const instance = getDOMNode(<Drawer.Actions classPrefix="custom-prefix" />);
-    assert.include(instance.className, 'custom-prefix');
+    const { container } = render(<Drawer.Actions>{title}</Drawer.Actions>);
+    expect(container.firstChild).to.have.text(title);
```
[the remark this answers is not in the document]
restartgr: sorry i missing this test

SevenOutman (approved):
LGTM

PR #3208, opened 2023-05-22
feat(RadioTile): add support for RadioTile

simonguo:
When more description is needed for radio buttons in a form, the `RadioTile` component is required. It can add icons and descriptions based on the Radio component.

image

image


SevenOutman:
Do we need a separate `RadioTileGroup` component while we already have `RadioGroup` that behaves the same (and doesn't look so different)?

simonguo:
> Do we need a separate `RadioTileGroup` component while we already have `RadioGroup` that behaves the same (and doesn't look so different)?
> 
> ```js
> <RadioGroup>
>   <RadioTile />
>   <RadioTile />
>   <RadioTile />
> </RadioGroup>
> ```

The main difference is that `RadioTileGroup` uses `Stack` for layout, making spacing and alignment more flexible. Additionally, properties like `appearance` inside `RadioGroup` are meaningless for `RadioTile`. If `RadioTileGroup` and `RadioGroup` are merged together, it could lead to less maintainable code.

SevenOutman:
> The main difference is that `RadioTileGroup` uses `Stack` for layout, making spacing and alignment more flexible. Additionally, properties like `appearance` inside `RadioGroup` are meaningless for `RadioTile`. If `RadioTileGroup` and `RadioGroup` are merged together, it could lead to less maintainable code.

That makes fair sense.

The current implementation does not resemble [design material](https://www.figma.com/file/uEq9QBplcKUYZrcWWA3RK2/NXT-UI?type=design&node-id=6417-17670&t=LigyAsIwKHDmjbcd-4) as perfectly. How do you think they should be synced? Whether the implementation or the design material should be updated?

simonguo:
> The current implementation does not resemble [design material](https://www.figma.com/file/uEq9QBplcKUYZrcWWA3RK2/NXT-UI?type=design&node-id=6417-17670&t=LigyAsIwKHDmjbcd-4) as perfectly. How do you think they should be synced? Whether the implementation or the design material should be updated?

@SevenOutman  Components have been updated according to the latest design material.



https://rsuitejs.com/resources/design/
@LeightonYao  We need to update the Figma design materials here with an independent PR.

src/RadioTile/RadioTile.tsx
```diff
@@ -0,0 +1,133 @@
...
+  icon?: React.ReactNode;
+
+  /** Name to use for form */
+  name?: string;
+
+  /** Value, corresponding to the value of the RadioTileGroup, to determine whether the */
+  value?: T;
+
+  /** Callback function with value changed */
+  onChange?: (value?: T, event?: React.ChangeEvent<HTMLInputElement>) => void;
```
SevenOutman, line 39: Radio components has a `checked` argument for `onChange` callback. Is is intentionally omitted on RadioTile?

SevenOutman (commented):
Are RadioTile and RadioTileGroup supposed to support `readOnly` and `plaintext`?

simonguo:
> Are RadioTile and RadioTileGroup supposed to support `readOnly` and `plaintext`?

I'm not sure in what scenarios the `readOnly` and `plaintext` states will be used, so we can leave them out for now in this version. We can always add them in later if needed.

src/RadioTile/RadioTile.tsx
simonguo: The `checked` argument only has meaning in the `onChange` event handler of a `Checkbox`. In `RadioTile`, its value is always true.  I wish to remove it.

SevenOutman: That makes sense.

SevenOutman:
The ARIA `radio` does not seem to have correct label and description as [APG specifies](https://www.w3.org/WAI/ARIA/apg/patterns/radio/#roles_states_properties).

> - The radiogroup element has a visible label referenced by [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) or has a label specified with [aria-label](https://w3c.github.io/aria/#aria-label).
> - If elements providing additional information about either the radio group or each radio button are present, those elements are referenced by the radiogroup element or radio elements with the [aria-describedby](https://w3c.github.io/aria/#aria-describedby) property.

For example, in the screenshot below, the ARIA `radio` should have had label as "Private" and description as "Project access must be granted explicitly to each user. If this project is part of a group, access will be granted to members of the group.", but it turns out using the full content as label.

image

SevenOutman (changes requested):
The ARIA `radio` produced by RadioTile does not seem to have correct label and description as [APG specifies](https://www.w3.org/WAI/ARIA/apg/patterns/radio/#roles_states_properties).

> - The radiogroup element has a visible label referenced by [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) or has a label specified with [aria-label](https://w3c.github.io/aria/#aria-label).
> - If elements providing additional information about either the radio group or each radio button are present, those elements are referenced by the radiogroup element or radio elements with the [aria-describedby](https://w3c.github.io/aria/#aria-describedby) property.

For example, in the screenshot below, the ARIA `radio` should have had label as "Private" and description as "Project access must be granted explicitly to each user. If this project is part of a group, access will be granted to members of the group.", but it turns out using the full content as label.

image

simonguo:
> The ARIA radio produced by RadioTile does not seem to have correct label and description as [APG specifies](https://www.w3.org/WAI/ARIA/apg/patterns/radio/#roles_states_properties).

Done

SevenOutman (approved):

PR #3215, opened 2023-05-25
fix(Radio,Checkbox): remove leading space for Radio/Checkbox in groups

SevenOutman:
Before

image

After

image


simonguo (approved):

simonguo:
image
There is a test case that needs to be updated.

SevenOutman:
> There is a test case that needs to be updated.

It's updated now.

PR #3228, opened 2023-06-02
feat: Picker's onOpen & onClose should decouple with onEnter & onExit

MarvelSQ:
## Question

In the current picker, users cannot control the `open` prop with `onOpen` and `onClose` because these props are bound to the animation. As a result, they cannot be triggered before the animation.

## Changes

in this PR, `onOpen` and `onClose` will be separate from animation, these callback bound to inner open state change. allow user to control picker with `open` props

```tsx
```

### DIFF
--- a/src/Overlay/OverlayTrigger.tsx
+++ b/src/Overlay/OverlayTrigger.tsx
@@ -197,6 +197,7 @@ const OverlayTrigger = React.forwardRef(
       onContextMenu,
       onFocus,
       onBlur,
+      onOpen,
       onClose,
       onExited,
       ...rest
@@ -245,8 +246,9 @@ const OverlayTrigger = React.forwardRef(
         }
 
         setOpen(true);
+        onOpen?.();
       },
-      [delayOpen, setOpen]
+      [delayOpen, setOpen, onOpen]
     );
 
     const handleClose = useCallback(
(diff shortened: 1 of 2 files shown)


simonguo:
@MarvelSQ  What breaking changes might occur to users?

MarvelSQ:
> @MarvelSQ What breaking changes might occur to users?

`onOpen` & `onClose` will fired before animation. but in present pickers, these event are bound with `onEntered` & `onExited`

As a result, the events are triggered earlier than what the user expected.

simonguo:
@MarvelSQ  I think it's necessary to decouple `onOpen` and `onClose` from the animation state. Do you think we can add related test cases for the controlled `open` prop?

MarvelSQ:
@simonguo test case for open with onOpen/onClose added in CheckPicker. should every picker be given a test case?

simonguo:
> should every picker be given a test case?

I think it is possible to add tests for OverlayTrigger or PickerToggleTrigger.

MarvelSQ:
@simonguo OverlayTrigger test case added

simonguo (approved):

simonguo (approved):
I added tests for the open state of all pickers and the tests passed.

> https://github.com/rsuite/rsuite/pull/3228/files#diff-b21b43be905c048c4e9b302049b2fcf681c8e10bdd3a928b77fbb940134fb820R131-R219

Issue #3250, opened 2023-06-16
The default value of size in modal is size = "md", but by using discovery, the real default value is size = "sm"


wangwc-dev:
![image](https://github.com/rsuite/rsuite/assets/78397810/01db478e-0ec3-4612-9605-f91251b73b0f)

SevenOutman:
@simonguo The default `size` was changed from `"md"` to `"sm"` in this commit https://github.com/rsuite/rsuite/commit/442d89d448dd587af22ffe021aa25a2933fda1a2#diff-f28a9fd5ed6ea7217b41bae1832a5bf0013c6ed4acd850a9a018f14b31338d30R75 5 years ago. Was is an intentional API change?

simonguo:
> Was is an intentional API change?

I no longer remember the reason for the change here. Use the default value of `size` as `sm`. We'd better fix the incorrect defaults in the docs

PR #3265, opened 2023-06-26
fix(Modal): fixed default size in modal component

Babrov:
**Changes**

- Changed modal default size prop to `md`

**Issue**
#3250


SevenOutman:
Hi @Babrov, thanks for contributing this edit! However please let me figure out whether we intentionally chose `"sm"` as the default value of `size` prop first.

PR #3270, opened 2023-06-30
feat(InputPicker): add shouldDisplayCreateOption prop

### LINKED ISSUE #3267: Specify when should InputPicker display "Create option" action

Most of the time when we use a creatable InputPicker, we don't want user to create a new option with duplicated name with existing entries.

image

For now InputPicker hides "Create option" action when the search keyword matches any entry's `[valueKey]`, which is not ideal IMO. Instead we tend to avoid duplicated `[labelKey]` most of the time.

https://github.com/rsuite/rsuite/blob/fafef7ca40e09581ba810858365a3a74fcd1a9ec/src/InputPicker/InputPicker.tsx#L671-L673

Provide a new prop for users to specify whether to display "Create option" action with given keyword. Something like:

```typescript
type InputPickerProps = {
  shouldDisplayCreateOption?: (keyword: string, data: T[]) => boolean;
}
```

Usage example:

```jsx
<InputPicker
  shouldDisplayCreateOption={(keyword, data) => {
    return data.every(item => item.name !== keyword)
  }}
/>
```

SevenOutman:
## New `shouldDisplayCreateOption` prop of InputPicker component

Specify when to display/hide "Create option" action with given search keyword in a creatable InputPicker.

Example: A creatable InputPicker, but only display "Create option" action when no item's `name` matches searchKeyword

```jsx
<InputPicker
  data={data}
  creatable
  shouldDisplayCreateOption={(searchKeyword, filteredData) =>
    filteredData.every(item => item.name !== searchKeyword)
  )}
/>
```

### Signature

The `shouldDisplayCreateOption` function takes two arguments and returns a boolean value

```typescript
type InputPickerProps = {
  /**
   * @param searchKeyword  Value of the textbox
   * @param filteredData   The items filtered by the searchKeyword
   *
   * @returns Whether to display "Create option" action
   */
  shouldDisplayCreateOption?: (
    searchKeyword: string,
    filteredData: InputItemDataType[]
  ) => boolean;
}
```

---

Close #3267

### DIFF
--- a/src/InputPicker/InputPicker.tsx
+++ b/src/InputPicker/InputPicker.tsx
@@ -103,6 +103,20 @@ export interface InputPickerProps<T = ValueType>
 
   /** Called when the option is created */
   onCreate?: (value: ValueType, item: ItemDataType, event: React.SyntheticEvent) => void;
+
+  /**
+   * Customize whether to display "Create option" action with given textbox value
+   *
+   * By default, InputPicker hides "Create option" action when textbox value matches any filtered item's [valueKey] property
+   *
+   * @param searchKeyword Value of the textbox
+   * @param filteredData The items filtered by the searchKeyword
+   */
+  shouldDisplayCreateOption?: (
+    searchKeyword: string,
+    // FIXME-Doma Use generic type
+    filteredData: InputItemDataType[]
+  ) => boolean;
 }
 
 const InputPicker: PickerComponent<InputPickerProps> = React.forwardRef(
@@ -132,6 +146,7 @@ const InputPicker: PickerComponent<InputPickerProps> = React.forwardRef(
       menuAutoWidth = true,
       menuMaxHeight = 320,
       creatable,
+      shouldDisplayCreateOption,
       value: valueProp,
       valueKey = 'value',
       virtualized,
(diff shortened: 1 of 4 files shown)


src/InputPicker/InputPicker.tsx
```diff
@@ -668,7 +683,12 @@ const InputPicker: PickerComponent<InputPickerProps> = React.forwardRef(
 
       let items: ItemDataType[] = filterNodesOfTree(getAllData(), checkShouldDisplay);
 
-      if (creatable && searchKeyword && !items.find(item => item[valueKey] === searchKeyword)) {
+      if (
+        creatable &&
+        (typeof shouldDisplayCreateOption === 'function'
+          ? shouldDisplayCreateOption(searchKeyword, items)
+          : searchKeyword && !items.find(item => item[valueKey] === searchKeyword))
```
simonguo, line 690:
```diff
- items.find(item => item[valueKey] === searchKeyword)
+ items.find(item => item[valueKey] === searchKeyword || item[labelKey] === searchKeyword)
```
If adding a matching check of [labelKey] can solve the problem in the current scenario?

SevenOutman: This would break the current behavior, where users might expect not hiding items with a matching `labelKey`.

simonguo (approved):

PR #3286, opened 2023-07-12
fix(OverlayTrigger): fix handleClose will trigger twice

myNameIsDu:
https://github.com/rsuite/rsuite/assets/53305259/850ce8a5-7884-4a9a-bb60-fe11fe48302a

Sometimes Chrome will trigger `mouseover`  => `mouseout`  =>  `mouseover`, Because there's a hidden bug in `OverlayTrigger `,  and then it will cause strange problems similar to the one mentioned above.

### DIFF
--- a/src/Overlay/OverlayTrigger.tsx
+++ b/src/Overlay/OverlayTrigger.tsx
@@ -327,7 +327,7 @@ const OverlayTrigger = React.forwardRef(
 
     const handleDelayedClose = useCallback(() => {
       if (!enterable) {
-        handleClose();
+        return handleClose();
       }
 
       isOnTrigger.current = false;


SevenOutman:
@myNameIsDu could you provide a CodeSandbox demo? You may use this template https://codesandbox.io/s/rsuite-template-jdydg7

myNameIsDu:
> @myNameIsDu could you provide a CodeSandbox demo? You may use this template https://codesandbox.io/s/rsuite-template-jdydg7

sorry, I couldn't, because it depends on Chrome and I don't know how to trigger it

SevenOutman:
> sorry, I couldn't, because it depends on Chrome and I don't know how to trigger it

Then how do you know how to fix it? 🤔️

myNameIsDu:
> Then how did you know how to fix it? 🤔️

According to the triggering method in the video above, I have logged the called steps below

image

image

and then we can find the key is "handleDelayedClose" method doesn't return early.

simonguo (approved):

PR #3298, opened 2023-07-20
fix(SelectPicker): wrong padding for grouped options

SevenOutman:
Before fix
image

After fix
image

Design
image


simonguo (approved):
`CheckPicker` has the same problem.

SevenOutman:
> `CheckPicker` has the same problem.

I'll try fixing it in a separate PR.

LeightonYao:
cool

PR #3306, opened 2023-07-25
fix(CheckPicker): wrong padding of grouped options

SevenOutman:
Before fix

image

After fix

image

Design

image


simonguo (approved):

LeightonYao:
Cool

PR #3323, opened 2023-08-08
fix(Picker): change color of caret on option group title

SevenOutman:
The caret color should be `@B600` per design, but was implemented as `@B800`.

Design

image

Before fix

image

After fix

image


simonguo (approved):

LeightonYao:
OK

PR #3324, opened 2023-08-08
fix(InputGroup): fix inside addon overlapping input text

SevenOutman:
## What's fixed

This edit fix 2 spacing issues with `<InputGroup inside>` addons

- The addon overlaps with input text when its content is long
- Gap between the addon content and input text was 13px, which should be 12px per design

Before

image

After

image

```jsx
<InputGroup inside>
  <InputGroup.Addon>https://</InputGroup.Addon>
  <Input defaultValue="www." />
</InputGroup>
```

As a side effect, this edit also fixes the doubled focus ring when InputGroup has focus

Before

image

After

image

## How it's done

The InputGroup with inside addons used to be implemented by setting `position: absolute` on the addon, making the Input the same size of the InputGroup, then increase the horizontal padding of Input to make room for the addon.

image

This caused the overlapping because when addon content increases the padding of Input doesn't change.

image

It's fixed by removing `position: absolute` from the addon and remove left padding of Input.

image

As a result, the Input's border-box does not fill the InputGroup, and the double focus ring problem was revealed.

image

It's fixed by removing the focus ring from Input when it's within an InputGroup.

### DIFF
--- a/docs/pages/components/input/fragments/input-group-inside.md
+++ b/docs/pages/components/input/fragments/input-group-inside.md
@@ -30,6 +30,11 @@ const App = () => (
     </InputGroup>
 
     <InputGroup inside style={styles}>
+      <InputGroup.Addon>https://</InputGroup.Addon>
+      <Input defaultValue="www." />
+    </InputGroup>
+
+    <InputGroup inside style={styles}>
       <Input />
       <InputGroup.Addon>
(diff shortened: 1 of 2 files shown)


simonguo (approved):

LeightonYao:
wow, it's cool

PR #3325, opened 2023-08-08
fix(DatePicker,DateRangePicker): fix calendar styles per design

SevenOutman:
### Month selection button overlapped

The month buttons should have 2px horizontal gap but overlapped.

Before

image

After

image

### Divider of month selection panel in dashed style

It was dotted but should be dashed per design.

Before

image

After

image

### Week number column shrank

The week number column should be 30px wide but shrank.

Before

image

After

image

## How it's done

The overlapping of month selection buttons was related to design issues. The width of the calendar panel was 246px, which could not contain the month selection buttons with appropriate gap. It's increased to 264px now. As a result, the horizontal gap between date buttons on the calendar panel (without week numbers) is increased from 2px to 4px, and horizontal padding of the calendar panel (without week numbers) is increased from 12px to 15px.

Design before vs. after

image

The week number column shrank because it's not given a correct width value (assigned `width: 1%`). As the calendar is implemented with `display: table`, adding `width: 30px` to the cell does not take effect. A quick fix is to assign `min-width: 30px`, but I think it's better reimplement it with grid layout.

### DIFF
--- a/src/Calendar/styles/index.less
+++ b/src/Calendar/styles/index.less
@@ -346,6 +346,7 @@
 }
 
 // Calendar Table
+// TODO-Doma Use CSS grid layout.
 .rs-calendar-table {
   display: table;
   table-layout: fixed;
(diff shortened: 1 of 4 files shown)


simonguo (approved):

LeightonYao:
yeah, great~

Issue #3245, opened 2023-06-14
InputPicker with virtualization select elements only after double clicks


sergeythrees:
### What version of rsuite are you using?

5.35.1

### What version of React are you using?

17.0.2

### What version of TypeScript are you using (if any)?

5.0.3

### What browser are you using?

Chrome

### Describe the Bug

InputPicker with virtualization item selects only after double clicks. 

[Form-Builder (3).webm](https://github.com/rsuite/rsuite/assets/22077517/f468b4dc-458a-4fd6-b918-2cbd29a90050)

### Expected Behavior

Without virtualization item selects by single click.
In SelectPicker with virtualization it works correct.

### To Reproduce

```js
  const listProps: Partial<ListProps> = {
    onItemsRendered: ({visibleStopIndex}) => {
      if (visibleStopIndex >= data.length - 1) {
        load()
      }
    }
  }

  const renderMenu = (menu: ReactNode) => <>
    {menu}
    {loading && <Loader/>}
  </>

  return (
      <InputPicker {...props} data={data} listProps={listProps} renderMenu={renderMenu} virtualized/>
  )
```

SevenOutman:
I'm routing this issue to @simonguo as it's related to virtualization.

shu8:
I'm also facing this issue with a `virtualized` `TagPicker`.

I found that the double click doesn't seem to actually need to be on the _same_ element. e.g., if you click on `element 1` then `element 2`, then `element 2` is what is finally checked. In fact it seems like the first click can be anywhere within the dropdown menu (e.g., any white space inside it).

This makes me wonder if the first click is acting as some sort of 'focus' event first?

I tried digging into the code but couldn't figure out exactly what's happening unfortunately. It seems like internally, `TagPicker` uses `InputPicker` which uses `DropdownMenu`. When I try a plain `DropdownMenu` that is `virtualized`, I don't need to double click. But when I use a `TagPicker`, I do need to double click.

Works (no need for double click):
```
<DropdownMenu
      dropdownMenuItemAs={DropdownMenuCheckItem}
      classPrefix="test"
      data={[{ label: "test", value: "test" }]}
      virtualized
      onSelect={(e) => console.log(e)}
/>
```

Doesn't work (requires double click):
```
<TagPicker
      data={[{ label: "test", value: "test" }]}
      virtualized
      onSelect={(e) => console.log(e)}
/>
```

[...]

PR #3329, opened 2023-08-09
Fix InputPicker select bug

sergeythrees:
Fix: #3245

InputPicker with virtualization item selects only after double clicks.

Calling set Open(false) in onBlur is incorrect because InputPicker onBlur is also called when clicking on InputPicker items. Closing InputPicker works without this function, when clicking outside the component.

For reproduce:
```
import React from "react";
import ReactDOM from "react-dom";
import { InputPicker, Loader } from 'rsuite';
import "rsuite/dist/rsuite.min.css";
import "./styles.css";

const fetchData = (start, length) => {
  return Array.from({ length }).map((_, index) => {
    return {
      label: `Item ${start + index}`,
      value: `Item ${start + index}`
    };
  });
};

const FixedLoader = () => (
  <Loader
    content="Loading..."
    style={{
      display: 'flex',
      justifyContent: 'center',
      position: 'absolute',
      bottom: '0',
      background: '#fff',
      width: '100%',
      padding: '4px 0'
    }}
  />
);

const App = () => {
  const [data, setData] = React.useState(fetchData(0, 30));
  const [loading, setLoading] = React.useState(false);

  const loadMore = () => {
    setLoading(true);
    setTimeout(() => {
      setData([...data, ...fetchData(data.length, 30)]);
      setLoading(false);
    }, 1000);
  };

  const onItemsRendered = props => {
    if (props.visibleStopIndex >= data.length - 1) {
      loadMore();
    }
  };

  const renderMenu = menu => {
    return (
      <>
        {menu}
        {loading && <FixedLoader />}
      </>
    );
  };

  return (
    <InputPicker
      data={data}
      style={{ width: 224 }}
      virtualized
      renderMenu={renderMenu}
      listProps={{
        onItemsRendered
      }}
    />
  );
};

ReactDOM.render(<App />, document.getElementById('root'));
```

### DIFF
--- a/src/InputPicker/InputPicker.tsx
+++ b/src/InputPicker/InputPicker.tsx
@@ -594,10 +594,6 @@ const InputPicker: PickerComponent<InputPickerProps> = React.forwardRef(
       }
     }, [readOnly]);
 
-    const handleBlur = useCallback(() => {
-      setOpen(false);
-    }, []);
-
     const handleEnter = useCallback(() => {
       focusInput();
       setOpen(true);
@@ -849,7 +845,7 @@ const InputPicker: PickerComponent<InputPickerProps> = React.forwardRef(
                   {...inputProps}
                   tabIndex={tabIndex}
                   readOnly={readOnly}
-                  onBlur={createChainedFunction(handleBlur, onBlur)}
+                  onBlur={onBlur}
                   onFocus={createChainedFunction(handleFocus, onFocus)}
                   inputRef={inputRef}
                   onChange={handleSearch}


src/InputPicker/InputPicker.tsx
```diff
@@ -594,10 +594,6 @@ const InputPicker: PickerComponent<InputPickerProps> = React.forwardRef(
       }
     }, [readOnly]);
 
-    const handleBlur = useCallback(() => {
-      setOpen(false);
```
sergeythrees, line 598: Calling set Open(false) in onBlur is incorrect because Dropdown onBlur is also called when clicking on Dropdown items. Closing Dropdown works without this function, when clicking outside the component.

simonguo (approved):

SevenOutman:
Thanks!

PR #3337, opened 2023-08-16
feat(Next.js): add 'use client'

psebestyen:
Added 'use client' to index.tsx to make components compatible with Next.js 13. Next.js requires the 'use client' directive for all components with client-side interactivity before they can be directly imported into server components. A more thorough approach would be to review all components to see if they can be used as server components and therefore not requiring the 'use client' directive, but that is realistically infeasible. Also, adding this only to index.tsx does not work with default imports, only with named imports (e.g. import { Button } from 'rsuite').

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,3 +1,5 @@
+'use client';
+
 // Base
 export { default as Button } from './Button';
 export type { ButtonProps } from './Button';


SevenOutman:
@psebestyen Thanks for contributing this edit!

Could you provide a working demo of Next.js 13 (RSC) using rsuite with this change?
You can install the package built from this PR by CodeSandbox with:

```shell
npm i https://pkg.csb.dev/rsuite/rsuite/commit/e91a2158/rsuite
```

simonguo (commented):
Hi @psebestyen 

Developers are unable to distinguish the difference between the following two methods in Next.js, which needs to be explained in the documentation.

```js
import { Button } from 'rsuite';
import Button from 'rsuite/Button';
```

My suggestion is to add `use-client` to the index.tsx file of all components.

PR #3355, opened 2023-08-31
feat(Next.js): add 'use client' to all components

psebestyen:
Added 'use client' to the index of all components to make them compatible with Next.js 13

### DIFF
--- a/src/Affix/index.tsx
+++ b/src/Affix/index.tsx
@@ -1,3 +1,4 @@
+'use client';
 import Affix from './Affix';
 import type { AffixProps } from './Affix';
 
--- a/src/Animation/index.tsx
+++ b/src/Animation/index.tsx
@@ -1,3 +1,4 @@
+'use client';
 import Transition, { TransitionProps } from './Transition';
 import Slide, { SlideProps } from './Slide';
 import Collapse, { CollapseProps } from './Collapse';
--- a/src/AutoComplete/index.tsx
+++ b/src/AutoComplete/index.tsx
@@ -1,3 +1,4 @@
+'use client';
 import AutoComplete from './AutoComplete';
 export type { AutoCompleteProps } from './AutoComplete';
 export default AutoComplete;
--- a/src/Avatar/index.tsx
+++ b/src/Avatar/index.tsx
@@ -1,3 +1,4 @@
+'use client';
 import Avatar from './Avatar';
 export type { AvatarProps } from './Avatar';
 export default Avatar;
--- a/src/AvatarGroup/index.tsx
+++ b/src/AvatarGroup/index.tsx
@@ -1,3 +1,4 @@
+'use client';
 import AvatarGroup from './AvatarGroup';
 export type { AvatarGroupProps } from './AvatarGroup';
 export default AvatarGroup;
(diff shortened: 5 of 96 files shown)


SevenOutman:
@psebestyen Thanks for contributing this edit!

Could you provide a working demo of Next.js 13 (RSC) using rsuite with this change?
You can install the package built from this PR by CodeSandbox with:

```shell
npm i https://pkg.csb.dev/rsuite/rsuite/commit/b9544eca/rsuite
```

psebestyen:
Sorry for the delay with this. As I was trying to create a sample Nextjs 13 app I ran into some issues. Specifically, with nested components such as Nav.Item and Nav.Menu. The error I am getting is 

> Error: Unsupported Server Component type: undefined

This is the case even if I add 'use client' at the top of each individual component, not just to index.tsx. There is no obvious solution as far as I can tell but I don't know enough about Nextjs to be sure. Does anyone have any suggestions?

By the way, creating a sample Nextjs app is really simple with npx create-next-app@latest (use the app directory option) in case anybody wants to try.

iska9der:
> Sorry for the delay with this. As I was trying to create a sample Nextjs 13 app I ran into some issues. Specifically, with nested components such as Nav.Item and Nav.Menu. The error I am getting is
> 
> > Error: Unsupported Server Component type: undefined

Try to add `'use client'` to `src/index.tsx`, maybe that would help

iska9der:
@psebestyen 
After install package from this PR by [command](https://github.com/rsuite/rsuite/pull/3355#issuecomment-1702285565), 
and insert `'use client'` directly to `esm/index.js`, 
[Error: Unsupported Server Component type: undefined](https://github.com/rsuite/rsuite/pull/3355#issuecomment-1708278474)
is dissapeared, but im facing next issue
image

Found related issues:
[nextjs issue](https://github.com/vercel/next.js/issues/51593)
[antd faq](https://ant-design.antgroup.com/docs/react/faq#the-antd-component-reported-an-error-when-using-the-app-router-of-nextjs):

> At present, this problem is waiting for Next.js to give an official solution, before this, if you use sub-components in your page, you can try to add the following client tag at the top of the page to solve this problem

MyNameIsTakenOMG:
Interesting, I didn't get any of this `sub-components` type of issue as @iska9der did when testing it in a code sandbox, or maybe I missed out something?🤔 Anyways, here are some screenshots below: 
Capture1
Capture2


As you can see, the version of `rsuite` that I got is from here: 

> @psebestyen Thanks for contributing this edit!
> 
> Could you provide a working demo of Next.js 13 (RSC) using rsuite with this change? You can install the package built from this PR by CodeSandbox with:
> 
> ```shell
> npm i https://pkg.csb.dev/rsuite/rsuite/commit/b9544eca/rsuite
> ```

psebestyen:
If you have 'use client' at the top everything works fine. The idea was to be able to embed rsuite components in react server components, that is, no 'use client' at the top. To achieve this, I tried adding 'use client' in the rsuite components themselves (or index files) which works for most components (meaning I could embed them in server componets) but I was not able to do the same with these sub-components.

simonguo:
@psebestyen  I will use a script to add 'use client' to each component. 

https://github.com/rsuite/rsuite/pull/3407

Issue #3248, opened 2023-06-15
'findDOMNode' is not exported from 'react-dom' (imported as 'findDOMNode').


megetron:
### What version of rsuite are you using?

5.35.1

### What version of React are you using?

18.2.0

### What version of TypeScript are you using (if any)?

5.1.3

### What browser are you using?

chrome

### Describe the Bug

when importing rsuite and using the TagGroup and Tag components, then i get error when close a tag (pressing on 'x'):
'findDOMNode' is not exported from 'react-dom' (imported as 'findDOMNode').

### Expected Behavior

no error should be displayed.

### To Reproduce

just install the above versions and create taggroup and closable tags. 
close a tag and get the error.

to fix the issue on my side i opened:
vi ./node_modules/.pnpm/rsuite@5.35.1_biqbaboplfbrettd7655fr4n2y/node_modules/rsuite/esm/utils/getDOMNode.js
import ReactDOM from ‘react-dom’;
instead of
import {ReactDOM} from ‘react-dom’;

SevenOutman:
Hi @megetron, I tried closable Tags in [this CodeSandbox](https://codesandbox.io/s/naughty-wind-vk7wr8?file=/src/index.js) couldn't re-produce the problem you described. Could you provide a CodeSandbox or a demo repo for reproduction?

minhktmd:
I have same problem with my nextjs project when I try using some rsuite components like (Loader, SelectPicker, Drawer)
It's a warning, not an error but it is so annoying

- warn ./node_modules/rsuite/esm/utils/getDOMNode.js
Attempted import error: 'findDOMNode' is not exported from 'react-dom' (imported as 'findDOMNode').

Import trace for requested module:
./node_modules/rsuite/esm/utils/getDOMNode.js
./node_modules/rsuite/esm/utils/index.js
./node_modules/rsuite/esm/Button/Button.js
./node_modules/rsuite/esm/Button/index.js
./node_modules/rsuite/esm/index.js

simonguo:
Hi  @megetron @minhktmd , I am unable to reproduce the issue locally, so I've made a preventative fix.
Please install a temporary version of rsuite using the following command and check if this resolves the issue.



```
npm i https://pkg.csb.dev/rsuite/rsuite/commit/4656cae1/rsuite
```

minhktmd:
Hi @simonguo 
I installed the temporary version of rsuite you give above and the problem has been resolved.
[...]

PR #3362, opened 2023-09-06
fix: add a safety check for the deprecated findDOMNode

simonguo:
fix https://github.com/rsuite/rsuite/issues/3248

### DIFF
--- a/src/utils/getDOMNode.ts
+++ b/src/utils/getDOMNode.ts
@@ -1,4 +1,12 @@
-import { findDOMNode } from 'react-dom';
+import ReactDOM from 'react-dom';
+
+function safeFindDOMNode(componentOrElement: React.Component | Element | null | undefined) {
+  if (componentOrElement && 'setState' in componentOrElement) {
+    // eslint-disable-next-line react/no-find-dom-node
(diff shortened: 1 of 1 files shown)


simonguo:
> What caused the issue and why this change fixes it?
Because there is no findDOMNode function in the latest react-dom. It should not be imported directly.

Because there is no findDOMNode function in the latest react-dom. It should not be imported as follows.

```diff
- import { findDOMNode } from 'react-dom';
```

SevenOutman:
> Because there is no findDOMNode function in the latest react-dom.

I wonder if it's the real cause.

It seems that `react-dom` is still exporting `findDOMNode`, per [latest source code](https://github.com/facebook/react/blob/ee7f9c9351f8902e07ceacf4234ef75e7e4ecd73/packages/react-dom/index.js#L17). And `import { findDOMNode } from 'react-dom'` also works correctly without warning in [this CodeSandbox](https://codesandbox.io/s/import-finddomnode-8qdm4p?file=/src/index.js) with latest `react-dom`.

simonguo:
I cannot reproduce this issue in my environment, and I suspect it is caused by their environment using some kind of verification tool.

PR #3392, opened 2023-10-11
fix(Navbar): fix background color of the subtle navigation items incorrect

simonguo:
## Before
image

## After 
image
> https://rsuitejs.com//design/default/index.html#s30


LeightonYao:
yeah~

Issue #3390, opened 2023-10-10
Blur check trigger form


forispeti:
### What version of rsuite are you using?

5.40

### What version of React are you using?

18.2

### What browser are you using?

Chrome

### Describe the Bug

If the check trigger in form is set to blur, it always triggers required on the blurred field. Tested on version 4 and it works perfectly there.

### Expected Behavior

To validate the input field correctly on blur.

### To Reproduce

Have a Form control field with a isRequired validation and it always triggers on blur.
![image](https://github.com/rsuite/rsuite/assets/99190634/d894bfea-d98a-4528-8306-64ac346bf0ed)

PR #3399, opened 2023-10-17
fix(form): form doesn't pass correct formValue to context

myNameIsDu:
fix: https://github.com/rsuite/rsuite/issues/3390

The reason is the external `formValue` was directly passed in context as the `formvalue`.

This will create a strong dependency on external `formValue`. if there's no external `formValue`. `Form.Control` component will always receive an empty `formValue`.

The solution is to pass `realFormValue` in the `Form` component as the `formValue`. 

However. there are some other further problems that are caused by this change: 

1. if neither `formDefautltValue` nor `formValue` props are passed in the `Form` component. React will throw an "uncontrolled to be controlled" error.  Although I think it's the right behaviour. But it's still a breaking change because both `formValue` and `formDefaultValue` weren't necessary before.

2. The second problem is after adding the `formDefaultValue` of the `Form ` component. React will throw  "component has both  `value` and `defaultValue`".  This is because we have passed these two props directly

image

Therefore, we have to design the `FormControl` component and make it handling value or defaultValue inside which is like `Form` component.

image

Through this change, we can make sure there's always value passed to the children component(depending on the first problem, there must be a ` formDefaultValue` or a `formValue` that contains the current field). However, this introduces another breaking change.

image

When using the `Form`'s `formDefaultValue` and the`FormContol`'s defaultValue, the `Form`'s `formDefaultValue` will be used.  This is the opposite behaviour from before

### DIFF
--- a/src/Form/Form.tsx
+++ b/src/Form/Form.tsx
@@ -385,7 +385,6 @@ const Form: FormComponent = React.forwardRef((props: FormProps, ref: React.Ref<F
     () => ({
       getCombinedModel,
       checkTrigger,
-      formDefaultValue,
       errorFromContext,
       readOnly,
       plaintext,
@@ -402,7 +401,6 @@ const Form: FormComponent = React.forwardRef((props: FormProps, ref: React.Ref<F
     [
       getCombinedModel,
       checkTrigger,
-      formDefaultValue,
       errorFromContext,
       readOnly,
       plaintext,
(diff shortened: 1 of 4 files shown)


simonguo:
@myNameIsDu  
The feature request to allow default values on FormControl to be overridden comes from here https://github.com/rsuite/rsuite/pull/3399.

Can we consider making a judgment in `FormControl`. If there is no value corresponding to `FormControl` in `formValue` in `Form`, and there is no value set in `FormControl`, the `defaultValue` on `FormControl` is allowed to be overwritten.

simonguo (approved):

Issue #3442, opened 2023-11-09
Accessibility Issue in SelectPicker


vijaysince89:
### What version of rsuite are you using?

5.42.0

### What version of React are you using?

16.14.0

### What browser are you using?

Chrome and Edge

### Describe the Bug

We are using JAWS screen reader to verify the Accessibility of SelectPicker component and its only reading the SelectPicker Combo Box and the value that is selected. It can not able to read the list of available values inside the combo box.

For example, after focusing the combo box (SelectPicker) then I click on the enter button in the keyboard to open the listbox that shows all the available items to be selected. when I using the keyboard up and down arrow, the Jaws screen reader is not reading the values.

### Expected Behavior

The expected behavior is, when we navigate through the list box inside the combo box using the up and down keyboard arrow buttons, the Jaws screen reader should read those values.

simonguo:
Hi @vijaysince89, please help to confirm whether the problem has been fixed.

https://rsuite-nextjs-git-improve-picker-accessibility-rsuite.vercel.app/components/select-picker/

vijaysince89:
@simonguo It's working perfectly fine now. Thank you.

PR #3463, opened 2023-11-21
fix: improve accessibility of picker related components

simonguo:
- Focus the selected element when selecting a value using the up and down keys on the keyboard.  fix: https://github.com/rsuite/rsuite/issues/3442
- Focus returns to trigger button after options list is closed.


vijaysince89:
Is this fix released ? If yes, which version?

simonguo:
@vijaysince89  Fixed in v5.46.0

PR #3486, opened 2023-12-04
refactor(Nav): remove legacy getDOMNode from tests

Nepsha1986:
According to [discussions/3200](https://github.com/rsuite/rsuite/discussions/3200) removed legacy getDomeNode from tests of "Nav" conponent.

"Nav" has no "role" attribute, so used  data-testid="nav" for querying.
"Nav.Item" has role "button", but in this case better querying with "getByText".

"Nav.Dropdown" changed to "Nav.Menu"  ("Nav.Dropdown" is depricated);
"Nav. Dropdown.Item"  => "Nav.Item" ("Nav. Dropdown.Item" is depricated);


simonguo (approved):
LGTM

PR #3496, opened 2023-12-08
refactor(Panel): remove legacy getDOMNode from tests

Nepsha1986:
I had to add /* eslint-disable testing-library/no-node-access */ for a whole file because we have a few places where the linter doesn't like it when we query the DOM node this way: screen.getByText(text).parentNode. Unfortunately, as we don't have any 'role' or other appropriate data attribute, currently, I do not see other options. If you have any ideas, please share.


simonguo (approved):
LGTM

PR #3497, opened 2023-12-12
refactor(Rate): remove legacy getDOMNode from tests

Nepsha1986:
I'm not sure that it is possible to avoid using "container" and node access for this tests. 
Had to add linter rules for the file, but it looks like that for other components the situation with querying will be the same - so  maybe it's not a bad idea to add this rules globally. 
/* eslint-disable testing-library/no-node-access, testing-library/no-container */
Screen without disabled linter rules:
Screenshot 2023-12-13 at 00 06 09

also switched tests from "ReactTestUtils" to "userEvents" where it was possible.

### DIFF
--- a/src/Rate/test/RateSpec.tsx
+++ b/src/Rate/test/RateSpec.tsx
@@ -1,7 +1,8 @@
+/* eslint-disable testing-library/no-node-access, testing-library/no-container */
 import React from 'react';
 import { render, act, screen } from '@testing-library/react';
 import ReactTestUtils from 'react-dom/test-utils';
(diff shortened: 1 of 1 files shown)


simonguo:
Hi @Nepsha1986 

In these test scripts, we try to follow the rules "Disallow direct Node access". If direct access to nodes cannot be avoided, we disable the rule in the file or line of code.

About `ReactTestUtils.Simulate` use `fireEvent` or `userEvents` to replace all

simonguo (commented):
LGTM

simonguo (approved):

PR #3511, opened 2023-12-18
FormControl multi-dimensional data support

yybawang:
The mysql json type in background development is becoming more and more common, and the one-dimensional data structure is difficult to meet the current business, so i write a wrapped FormControl component in my project like this do, if we can merge to origin that will he helpful..
(Non-breaking update, I wrote some test cases and tested the shouldResetWithUnmound configuration)

### DIFF
--- a/src/FormControl/FormControl.tsx
+++ b/src/FormControl/FormControl.tsx
@@ -1,6 +1,6 @@
 import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
-import isUndefined from 'lodash/isUndefined';
+import { isUndefined, get as lodashGet, set as lodashSet } from 'lodash';
 import Input from '../Input';
 import FormErrorMessage from '../FormErrorMessage';
 import { useClassNames } from '../utils';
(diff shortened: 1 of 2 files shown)


src/FormControl/test/FormControlSpec.tsx
```diff
@@ -41,6 +41,29 @@ describe('FormControl', () => {
     expect(onChange).to.have.been.calledOnce;
   });
 
+  it('Should set correctly defaultValue', () => {
+    // const onChange = sinon.spy();
```
simonguo, line 45: Remove unused `onChange`

src/FormControl/FormControl.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useContext } from 'react';
 import PropTypes from 'prop-types';
-import isUndefined from 'lodash/isUndefined';
+import { isUndefined, get as lodashGet, set as lodashSet } from 'lodash';
```
simonguo, line 3: Import lodash on demand to avoid excessive package size in projects that do not use Tree shaking.

```suggestion
import isUndefined from 'lodash/isUndefined';
import lodashGet from 'lodash/lodashGet';
import lodashSet from 'lodash/lodashSet';
```

simonguo (changes requested):

src/FormControl/FormControl.tsx
yybawang: > Import lodash on demand to avoid excessive package size in projects that do not use Tree shaking.

typo error:
`lodash/lodashGet` should fix to `lodash/get`，and lodashSet too

simonguo: Yes, it should be `lodash/get`

simonguo:
Hi @yybawang  do you have time to fix the above mentioned problems these days, and plan to release this Friday

yybawang:
Hi @simonguo , I submitted it again (sorry i just saw it now)

simonguo (approved):

PR #3561, opened 2024-01-16
fix: NavItem and DropDownItem rendering icon without overriding className

myNameIsDu:
In both NavItem and DropdownItem, the className attribute assigned to the icon prop will be overridden.

image

notice:  there's an extra test case change, but I think it was wrong before.

### DIFF
--- a/src/Dropdown/test/DropdownItemSpec.tsx
+++ b/src/Dropdown/test/DropdownItemSpec.tsx
@@ -82,12 +82,20 @@ describe('<Dropdown.Item>', () => {
 
   it('Should render a icon', () => {
     render(
-      <Dropdown>
-        <DropdownItem icon={<User data-testid="icon" />} data-testid="item" />
+      <Dropdown open>
+        <DropdownItem icon={<User data-testid="icon" />} />
       </Dropdown>
     );
 
-    expect(screen.getByTestId('item')).to.contain(screen.getByTestId('icon'));
+    expect(screen.getByRole('menuitem')).to.contain(screen.getByTestId('icon'));
+  });
+  it('Should render a icon without overriding className', () => {
+    render(
+      <Dropdown open>
+        <DropdownItem icon={<div className="custom-class" />} data-testid="item" />
+      </Dropdown>
+    );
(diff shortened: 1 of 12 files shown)


src/Dropdown/test/DropdownItemSpec.tsx
```diff
@@ -87,7 +87,18 @@ describe('<Dropdown.Item>', () => {
       </Dropdown>
     );
 
-    expect(screen.getByTestId('item')).to.contain(screen.getByTestId('icon'));
+    expect(screen.getByTestId('icon')).to.contain(screen.getByTestId('icon'));
```
simonguo, line 90: icon contains icon is logically incorrect, it is recommended to use getByRole('treeitem) instead.

simonguo (changes requested):
Missing tests for icon are in NavDropdownItem, NavbarDropdownItem, SidenavDropdownItem, SidenavItem components

src/Dropdown/test/DropdownItemSpec.tsx
myNameIsDu: Done, there's no need to make any modifications to this change.

simonguo: `Unable to find an accessible element with the role "menuitem"` error is caused by the DropdownItem not being visible, you need to set an open attribute for the Dropdown.
```js
it('Should render a icon without overriding className', () => {
    render(
      <Dropdown open>
        <DropdownItem icon={<div className="custom-class" />} />
      </Dropdown>
    );
    expect(screen.getByRole('menuitem')).to.contain('.custom-class');
  });
```

According to the testing-library's suggestion, we prefer to use `getByRole` to get the element.
https://testing-library.com/docs/queries/about#priority

myNameIsDu: done

myNameIsDu:
> Missing tests for icon are in NavDropdownItem, NavbarDropdownItem, SidenavDropdownItem, SidenavItem components

done

simonguo (approved):

PR #3774, opened 2024-04-30
fix(toggle): block pointer events when disabled

ShrinidhiUpadhyaya:
### Description
When the toggle is disabled it is still accepting pointer events.

https://github.com/rsuite/rsuite/assets/8769408/ca752fbe-b18c-465d-b3bc-d1b039c4c595

### Current behavior
`cursor: not-allowed;`

### New behavior
`pointer-events: none;`


simonguo (approved):
LGTM

Issue #3860, opened 2024-06-27
`@enable-high-contrast` variable is not actually supported


jasonbarry:
### What version of rsuite are you using?

5.46.2

### What version of React are you using?

18.3.0

### What version of TypeScript are you using (if any)?

5

### What browser are you using?

Safari

### Describe the Bug

[In the docs](https://rsuitejs.com/guide/official-themes/#1-import-style-files), it says that `@enable-high-contrast: true` will include high contrast themes, which is correct. However, setting `@enable-high-contrast: false` does not exclude those styles from the dist CSS output. 

### Expected Behavior

I would expect that setting `@enable-high-contrast: false` would prevent high-contrast theme from being bundled in dist assets, since setting `@enable-dark-mode: false` prevents dark mode styles from being included in the bundle.

PR #3861, opened 2024-06-27
fix(theme): add less variable to opt-out of high-contrast theme bundling

jasonbarry:
Fixes https://github.com/rsuite/rsuite/issues/3860


simonguo (approved):

jasonbarry:
Just checked a before/after of my main CSS bundle: 

- before change: 373.76 kB
- after change: 324.99 kB

PR #3878, opened 2024-07-12
feat(locales): add support for Catalan

rogeranyor:


simonguo (changes requested):
The code is not formatted by prettier, you can format the code with the following command:
  
```bash
npm run format
```

simonguo (approved):

Issue #3899, opened 2024-07-25
Support for the non-english decimal separator


electroheadfx:
### What problem does this feature solve?

Support to allow comma in rsuite input Number like the HTML version of input.
example : `0,5` -> show `0.5` in js

### What does the proposed API look like?

just add a property for exemple:
`<InputNumber comma value={0.5} />`
Default will be period:
`<InputNumber value={0.5} />`

simonguo:
Hi @electroheadfx 

I added a `decimalSeparator` prop to InputNumber to customize the decimal separator. 
Here is an example https://rsuite-nextjs-git-feat-inputnumber-decimalseparator-rsuite.vercel.app/components/input-number/#decimal-separator

electroheadfx:
Amazing man it works great! thanks

electroheadfx:
hey, is it possible to allow to write both '.' or ',' and apply the decimalSeparator feature you did ?

PR #3907, opened 2024-08-02
feat(InputNumber): add support for `decimalSeparator` prop

simonguo:
To solve this requirement  https://github.com/rsuite/rsuite/issues/3899

In other languages and cultures, the decimal point may not be a ".". The main decimal point symbols used in countries and regions using Arabic numerals are "," and ".". Chinese-speaking areas and most English-speaking areas use ".", but most other European countries and their former colonies use ",".

Add a `decimalSeparator` property to officially designate different symbols as separators for different countries/regions.
### Usage
```
<InputNumber decimalSeparator="," />
```

### DIFF
--- a/docs/pages/components/input-number/en-US/index.md
+++ b/docs/pages/components/input-number/en-US/index.md
@@ -20,6 +20,12 @@ An input component that can only enter numbers.
 
 <!--{include:`decimals.md`}-->
 
+### Decimal separator
+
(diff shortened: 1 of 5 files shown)


electroheadfx:
My client want to type number decimal with "." and "," because on Windows computer the number keyboard are point "."
but allow to french to type with "," on french Mac keyboard its ","
is it a elegant way to manage it ? else not I am forced to use string input with validation ? 
Thanks, Iam developping the Third app with rsuite for railway industry ;)

PR #3922, opened 2024-08-15
fix(MultiCascader): add missing cascade props of SearchView in MultiCascader

myNameIsDu:


simonguo (commented):
In `MultiCascadeTree`, the `cascade` prop is also missing, and I hope to add a unit test for this.

https://github.com/rsuite/rsuite/blob/main/src/MultiCascadeTree/MultiCascadeTree.tsx#L80-L89

PR #3977, opened 2024-09-25
fix(DateRangePicker): fix month switching with the “show one calendar” property

xTrogwaRx:


simonguo (commented):
Please describe in detail the exact issue you fixed. Also please add unit tests for the added code.

xTrogwaRx:
> Please describe in detail the exact issue you fixed. Also please add unit tests for the added code.

If you set the showOneCalendar property in the DateRangePicker component, when you try to select a date in the current month, the calendar switches to the next month.

Expected behavior: the calendar stays on the same month, since the user may need to select a date in the current or previous months.

I unfortunately have never written tests, but I'll try to read about tests and try to add them in my spare time.

simonguo:
Understood, this is indeed an issue that needs improvement

simonguo:
Fixed in v5.72.0

PR #4010, opened 2024-10-20
refactor: replaced `cursor: not-allowed` with `cursor: @cursor-disabled`

ShrinidhiUpadhyaya:
Replaced `cursor: not-allowed` with `cursor: @cursor-disabled` across the following components to maintain code consistency
1. Checkbox
2. RadioTile
3. Rate
4. Slider

Also removed it in the following components as the cursor was already disabled.
1. CheckTree
2. TagPicker
3. Tree 

**Note:-** I have replaced it only in the components which i could test, `cursor: not-allowed` is still present in multiple components.

### DIFF
--- a/src/RadioTile/styles/index.less
+++ b/src/RadioTile/styles/index.less
@@ -1,3 +1,5 @@
+@import '../../styles/common';
+
 .rs-radio-tile {
   border-radius: 6px;
   overflow: hidden;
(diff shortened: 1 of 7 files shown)


src/RadioTile/styles/index.less
```diff
@@ -53,7 +53,7 @@
   &-disabled,
   &-disabled &-content {
     color: var(--rs-text-disabled);
-    cursor: not-allowed;
+    cursor: @cursor-disabled;
```
simonguo, line 56: In RadioTile, the reference to public variables is missing.

image

simonguo (changes requested):

PR #4012, opened 2024-10-21
feat(CustomProvider): support setting the default value of the component props globally

simonguo:
Define global default values through the `components` prop of the CustomProvider.

```jsx
import { CustomProvider } from 'rsuite';

return (
  <CustomProvider
    components={{
      Button: { defaultProps: { appearance: 'primary', size: 'lg' } },
      Input: { defaultProps: { placeholder: 'Enter text', size: 'lg' } }
    }}
  >
    <App />
  </CustomProvider>
);
```
> https://rsuite-nextjs-9f4qmggjx-rsuite.vercel.app/components/custom-provider/#global-configuration-of-default-values-for-components

### Preview

```
npm i rsuite@5.74.0-beta.0
```


ab-pm:
This PR seems to have broken forwarding the `locale` prop through [components composed with `as`](https://www.rsuitejs.com/guide/composition/#code-as-code-prop) or similar. The code in

https://github.com/rsuite/rsuite/blob/edaa278e48f17068122ba84d222ead338e5272a4/src/CustomProvider/useCustom.ts#L89-L94

only checks the `componentName` of the forwarder, but not of the target. An example where I am facing this is
```jsx
<Form.Control
     …
     accepter={InputPicker}
     locale={{createOption: '…'}}
/>
```
where the `<InputPicker>` will not receive the `locale` prop, since the `componentName` passed to `useCustom` is `"FormControl"`, causing the `locale` to be dropped silently. Rendering `<InputPicker locale={…} />` directly works fine.

Not sure what the proper solution would be here. I guess the `componentLocale` should still be put right in `mergedProps`, even if conditionally overwritten afterwards? Probably the `componentProps` should not be destructured in https://github.com/rsuite/rsuite/blob/edaa278e48f17068122ba84d222ead338e5272a4/src/CustomProvider/useCustom.ts#L58 at all but the hook code should simply use `componentProps` in place of `restProps` everywhere. (Using `restProps` as a `useMemo` dependency doesn't work anyway).

PR #3998, opened 2024-10-15
build: bump date-fns 4.1.0

obdulia-losantos:
Based on https://github.com/rsuite/rsuite/discussions/3602

PR #4079, opened 2024-12-23
build: bump date-fns@4.1.0

obdulia-losantos:
Based on:
- https://github.com/rsuite/rsuite/discussions/3602
- https://github.com/rsuite/rsuite/pull/3998

### Breaking change

- [Starting from v3.0.0, date-fns no longer supports IE 11.](https://github.com/date-fns/date-fns/releases/tag/v3.0.0)


obdulia-losantos:
@simonguo, I did not include the docs/examples update. I can do that but it could imply to change or maybe remove the orphan `package-lock.json` files. Also deal with "rsuite" dependency.

simonguo:
> @simonguo, I did not include the docs/examples update. I can do that but it could imply to change or maybe remove the orphan `package-lock.json` files. Also deal with "rsuite" dependency.

Hi @obdulia-losantos  `examples` directory does not need to be updated, but the date-fns in the `docs` directory needs to be updated together, otherwise the document will compile an error.

https://github.com/rsuite/rsuite/blob/main/docs/package.json#L36

obdulia-losantos:
@simonguo The problem is this dependency: https://github.com/rsuite/rsuite/blob/main/docs/package.json#L54. 

RSuite 5.76.0 uses `date-fns@2.x.x`.

obdulia-losantos:
> In the development phase, rsuite depends on the `src` directory, and the rsuite version in the document will be updated after the new version is released.
> https://github.com/rsuite/rsuite/blob/main/docs/next.config.js#L180

I saw that but the issue is that the installed version dependencies are the one indicated in the `package.json` file, then `rsuite@5.76.0` 5 with `date-fns@2.29.1`. 
If you check `package-lock.json` file you'll see both entries now.

Anyway, I changed the files with `date-fns` import statements, now they work with both versions, 2 and 4. That will do the trick till you change to the new release.

simonguo (changes requested):
LGTM

simonguo (approved):
LGTM

PR #4110, opened 2025-01-06
Fix font-family

alex-kinokon:
~~It’s `-apple-system` with a beginning dash.~~

This does not work in Chrome.

```css
@font-face {
  font-family: Apple-System;
  src: local(-apple-system), local(BlinkMacSystemFont), local(system-ui);
}
```

### DIFF
--- a/src/styles/variables.less
+++ b/src/styles/variables.less
@@ -124,7 +124,7 @@
 // Typography
 // Font, line-height, and color for body text, headings, and more.
 
-@font-family-base:      -apple-system, Arial, Helvetica, 'PingFang SC', 'Hiragino Sans GB', 'Microsoft YaHei', STXihei, sans-serif;
+@font-family-base:      -apple-system, BlinkMacSystemFont, Arial, Helvetica, 'PingFang SC', 'Hiragino Sans GB', 'Microsoft YaHei', STXihei, sans-serif;
 
 @font-size-extra-large: 18px;
 @font-size-large:       16px;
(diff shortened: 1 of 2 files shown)


src/styles/variables.less
```diff
@@ -124,7 +124,7 @@
 // Typography
 // Font, line-height, and color for body text, headings, and more.
 
-@font-family-base:      Apple-System, Arial, Helvetica, 'PingFang SC', 'Hiragino Sans GB', 'Microsoft YaHei', STXihei, sans-serif;
+@font-family-base:      -apple-system, Arial, Helvetica, 'PingFang SC', 'Hiragino Sans GB', 'Microsoft YaHei', STXihei, sans-serif;
```
simonguo, line 127: It is a way to be compatible with old browsers, defining [Apple-System](https://github.com/rsuite/rsuite/blob/main/src/styles/typography.less#L121
).
```css
// font-face
@font-face {
  font-family: Apple-System;
  src: local(-apple-system), local(BlinkMacSystemFont), local(system-ui);
}
```


I think it can be changed to:
```less
@font-family-base: -apple-system , BlinkMacSystemFont, ......;
```

Also need to remove the definition of `Apple-System`.

simonguo (changes requested):

src/styles/variables.less
alex-kinokon: None of `local(-apple-system)`, `local(BlinkMacSystemFont)`, `local(system-ui)` works in Chrome.

simonguo (approved):

PR #4166, opened 2025-03-04
fix: 1 test in ImageSpec

abhidatta0:
Noticed that 1 test case in Image was failing.
Screenshot 2025-03-04 at 10 07 05 AM


abhidatta0:
@simonguo please check this

simonguo (approved):

Issue #4180, opened 2025-03-12
Slider and Slider Range Keep tooltip open


sahand-sn:
### What problem does this feature solve?

to show current amount of slider title, I want to keep the tooltip open.

### What does the proposed API look like?

a flag like keepTooltipOpen, when true, tooltip will not go away after hover.

PR #4186, opened 2025-03-14
feature for #4180: keepTooltipOpen

abhidatta0:
Fixes: #4180


abhidatta0:
@simonguo please check

simonguo (changes requested):
Hi @abhidatta0 Thank you for submitting the PR. We've identified a few issues:
- `keepTooltipOpen` should be a configuration option for both Slider and RangeSlider, with a default value of `false`
- It's preferable to handle `keepTooltipOpen` as a parameter in useDrag to manage the initial tooltip position internally, rather than adding an extra useEffect in the Handle component.
- Please enhance the unit tests. For instance, test the scenario where the tooltip can still open when `keepTooltipOpen` is set to `true` but `tooltip` is `false`.

Here's a development preview link: https://rsuite-nextjs-git-fork-abhidatta0-fix-issue4180-rsuite.vercel.app/components/slider/

simonguo (approved):

PR #4254, opened 2025-04-20
FEAT: Add prop to disable transition animations in the calendar component, datepicker, and its derivatives.

ruxwez:
This PR adds a new prop for all elements that use the `DatePicker` component, allowing users to disable the transition animation between `day`, `month`, and `time` views.

The prop is called `isAnimated`, it defaults to `true`, and setting it to `false` removes the transition animations.

The main reason I wanted to implement this feature is because when opening a DatePicker in month selector mode or a TimePicker, the animation appears and feels awkward and visually unappealing.

### TimePicker component
https://github.com/user-attachments/assets/fefc07a6-5a73-40e5-96dd-3dcef836f54f

### DatePicker component

https://github.com/user-attachments/assets/80a95d4b-c3fe-47f2-bb65-1a18c6926982

### Calendar component

https://github.com/user-attachments/assets/52e99838-8cde-47d6-a61e-43becfaf367f

In addition, I created some tests to make sure the prop works and added the changes to the documentation page.


simonguo:
Hi @ruxwez , I understand your requirement, but I think adding a prop to disable animation to the [CustomProvider](https://rsuitejs.com/components/custom-provider/) component would be a more appropriate choice. This is similar to the `disableRipple` property on the CustomProvider component, which disables the ripple effect.

Additionally, I hope that enabling/disabling animations can be implemented through CSS variables, allowing for more flexible control over animation effects. Some CSS refactoring has been done in [rsuite v6](https://github.com/rsuite/rsuite/discussions/4228), so it would be best to place this feature in the [next](https://github.com/rsuite/rsuite/tree/next) branch.

PR #4256, opened 2025-04-22
feat: add Codecov bundle analysis support

RulaKhaled:
Add Webpack bundle analysis integration to track bundle size changes over time and in PRs. Since the repository already uses Codecov for coverage reporting, this adds bundle size tracking with minimal configuration, this will help monitor bundle size changes and prevent unexpected increases in future PRs.


simonguo:
Hi @RulaKhaled 

We have added [size-limit](https://github.com/rsuite/rsuite/blob/next/package.json#L206) and its corresponding [workflow](https://github.com/rsuite/rsuite/blob/next/.github/workflows/size-limit.yml) to the next branch.

Issue #4154, opened 2025-02-16
Type Error with Form onChange in React 19.0.0


Eldar-X:
### What version of rsuite are you using?

5.77.1

### What version of React are you using?

19.0.0

### What version of TypeScript are you using (if any)?

5.7.3

### What browser are you using?

Chrome

### Describe the Bug

When using RSuite's Form component with React 19, receiving a TypeScript error when trying to use the `onChange` prop with `useState` setter function. The example in the documentation works fine, but fails with strictly typed state.

## Error Message
```typescript
Type 'Dispatch<SetStateAction<{ name: string; code: string; type: string; status: string; }>>' 
is not assignable to type '(formValue: Record<string, any>, event?: SyntheticEvent<Element, Event> | undefined) => void'.

Types of parameters 'value' and 'formValue' are incompatible.

Type 'Record<string, any>' is not assignable to type 'SetStateAction<{ name: string; code: string; type: string; status: string; }>'.

### Expected Behavior

The Form's onChange should work with strictly typed state setter functions, as shown in the documentation example.Possible Solution

### Possible Solution
The Form component's onChange type definition might need to be updated to better handle strict typing in React 19's setState. Consider updating the type definition to:

```typescript
onChange?: (formValue: T, event?: SyntheticEvent<Element, Event>) => void;
```

### To Reproduce

1- Create a new React 19 project with TypeScript
2- Install RSuite
3- Create a form component using RSuite Form
4- Use strictly typed state with useState

```typescript
type FormState = {
  name: string;
  code: string;
  type: string;
  status: string;
};

const [formValue, setFormValue] = useState<FormState>({
  name: '',
  code: '',
  type: '',
  status: ''
});

return (
  <Form fluid onChange={setFormValue} formValue={formValue}>
    {/* Form fields */}
  </Form>
);
```

PR #4280, opened 2025-05-10
fix(Form): fix onChange type allowing strictly typed setState

nandobutzke:
This PR updates the onChange prop type in the Form component by replacing the previous V = Record<string, any> constraint with a more flexible generic type M.

The goal is to allow the use of strongly-typed state management functions (e.g., setState from useState) without triggering TypeScript type errors.

Under the previous typing, TypeScript raised an error when attempting to pass a setState function directly to onChange with a strictly typed formValue, like:

```typescript
type FormState = { name: string; code: string; type: string; status: string };
const [formValue, setFormValue] = useState<FormState>(...);
<Form formValue={formValue} onChange={setFormValue} />
```

Related issue: #4154

### DIFF
--- a/src/Form/Form.tsx
+++ b/src/Form/Form.tsx
@@ -17,7 +17,7 @@ import useFormValue from './hooks/useFormValue';
 import useFormClassNames from './hooks/useFormClassNames';
 import useFormRef, { FormInstance, FormImperativeMethods } from './hooks/useFormRef';
 
-type FormValue = any
+type FormValue = any;
 
 export interface FormProps<V = Record<string, any>, M = any, E = { [P in keyof V]?: M }>
   extends WithAsProps,


src/Form/Form.tsx
```diff
@@ -99,7 +99,7 @@ export interface FormProps<V = Record<string, any>, M = any, E = { [P in keyof V
   /**
    * Callback fired when data changing
    */
-  onChange?: (formValue: V, event?: React.SyntheticEvent) => void;
+  onChange?: (formValue: M, event?: React.SyntheticEvent) => void;
```
simonguo, line 102:
```
FormProps<V = Record<string, any>, M = any, E = { [P in keyof V]?: M }>
```

Hi @nandobutzke, for FormProps, V, M, and E respectively represent Value, Message, and Error.

Issue #4291, opened 2025-05-21
Notification closes immediately when shown after previous one (duration=5000, interval=5100)


kryaksy:
### What version of rsuite are you using?

5.70.2

### What version of React are you using?

18.3.1

### What version of TypeScript are you using (if any)?

5.4.5

### What browser are you using?

Chrome

### Describe the Bug

When pushing a Notification every 5100ms with a duration of 5000ms, I expect the new notification to appear after the previous one is auto-closed. However, what happens is that every second notification is auto-closed immediately upon appearing — as if both the previous and the current one are being closed together.

### Expected Behavior

Each notification should remain visible for its full duration (5000ms). When a new notification is pushed after 5100ms, the previous one should already be closed, and the new one should stay visible for another 5000ms. No notification should disappear immediately upon appearing.

### To Reproduce

https://codesandbox.io/p/sandbox/rsuite-notification-autoclose-issue-9cqhgc

kryaksy:
Sorry for that, forgot to make it public. It's public now.

simonguo:
The 5000ms and 5100ms intervals are too close. The `duration` here doesn't include the animation effect time. You can set the setInterval time a bit longer, for example: 6000ms

kryaksy:
I used setInterval here just to simulate a real-world scenario. In an actual application, notifications may be triggered by various user actions or background events, and it’s entirely possible that one appears shortly (e.g., 5100ms) after the previous one.

In real usage, multiple notifications can be triggered in unpredictable sequences, and checking whether each one collides with the animation duration of another can quickly become impractical or overly complex. The expectation is that each notification should independently respect its full duration, regardless of when it is pushed.

simonguo:
Fixed in  rsuite@5.81.0

kryaksy:
I am not sure it is resolved since no notifications can be toasted since v5.76.1. Even the codesandbox examples doesn't work. Like this one:
[https://codesandbox.io/p/sandbox/9f9g23](https://codesandbox.io/p/sandbox/9f9g23)

simonguo:
When using the `useToaster` hook, you need to wrap your App with `CustomProvider`:

```tsx
import { CustomProvider } from 'rsuite';

render(
  <CustomProvider>
    <App />
  </CustomProvider>,
  document.getElementById('root')
);
```

PR #4294, opened 2025-05-23
fix(toaster): fix issue where delayed closing affects state

simonguo:
This pull request updates the `useMessages` function in `src/toaster/ToastContainer.tsx` to ensure state updates are based on the previous state, improving reliability when handling asynchronous updates.

State management improvements:

* Updated the `setMessages` call to use a functional state update in the `setTimeout` for clearing all messages. This ensures the state update is based on the most recent state.
* Updated the `setMessages` call to use a functional state update when filtering visible messages. This prevents potential issues caused by stale state references.


simonguo:
fix: https://github.com/rsuite/rsuite/issues/4291

PR #4356, opened 2025-07-20
(feat) Stat.Trend - add number type to trend indicator props

paul-prism:
Stat.Trend property indicator should take a number directly, which represents the trend direction: < 0 for down, > 0 for up, o for equal. This will allows the component to be created from source data (which almost alwasy consists of the current numeric value, and some previous value: see example below) rather than the user having to evaluate the tri-state string.

`
<Stat bordered>
  <Stat.Label>Revenue</Stat.Label>
  <HStack spacing={10}>
    <Stat.Value>{currentValue}</Stat.Value>
    <Stat.Trend indicator={currentValue - previousValue}>{Math.round(100*(currentValue - previousValue)/previousValue)}%</Stat.Trend>
  </HStack>
</Stat>
`

### DIFF
--- a/docs/pages/components/stat/en-US/index.md
+++ b/docs/pages/components/stat/en-US/index.md
@@ -86,12 +86,12 @@ Used to display statistical data with a title and its corresponding value, empha
 
 ### `<Stat.Trend>`
 
-| Property    | Type `(Default)`          | Description                           |
-| ----------- | ------------------------- | ------------------------------------- |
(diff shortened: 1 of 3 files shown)


simonguo:
Hi @paul-prism 

I don't suggest adding number type to `Stat.Trend` props because `Stat.Trend` is designed to show the trend change, not to show the specific numerical value. This will make the usage of Stat.Trend unclear and cause confusion for users when using it.

PR #4421, opened 2025-11-03
feat(Uploader): add `onAllUploadComplete` callback

### LINKED ISSUE #813: Function for when all files finish uploading.

Currently, the Uploader component works such that when a single image finishes uploading, you can latch onto the `onSuccess` method if individual requests. While this is useful, it's a lot less so when you have, say, ten different images that need to be uploaded to a third-party service and another function should be called once it's done.

I'm trying to build an application with ImageKit so that once all the images are uploaded, a function can be called and I can then send the aggregated responses at once to the server for processing, rather than one by one. (Sending them one at a time creates an additional problem since it's impossible to sort them out on the server without some sort of identifier). 

I'd imagine others would find this useful for letting the client know that the process was completed successfully or something like that.

My attempt at a solution has been tracking the number of files that have been selected by incrementing  a variable every time a new image is added (and decreasing that count if an image is removed), like this:

```js
state = {
    imageCount: 0,
    uploadCount; 0
}

//Every file added to the uploader is counted
handleChange(values){
    this.setState({
          ...this.state,
          imageCount: this.state.imageCount += values.length  
    })
} 

//Since this is called for every individual successful upload, we can just increment it by 1
handleUploadSuccess(){
     this.setState({
           ...this.state,
           uploadCount: this.state.uploadCount += 1    
      })
}

//... logic for removing count... etc.

```

This suffers from the drawback of being unable to 'listen' using a callback for once every file is uploaded, however, so that an extra button/function has to be added instead. eg.

```jsx
//...
render(){
    return(
        <div>
            {
                 this.state.imageCount === this.state.uploadCount &&
                 <div>
                         <p>Success message.</p>
                 </div>
             }
        </div>
    )
}
```
Essentially, it's impossible to have a dynamic function that listens for the completion of every upload/at the very least track the number of successful uploads and the failed ones (if any) . As you might imagine, that would also involve a lot more code.

Instead, the library might contain something like an `onCompletion` prop so that

```js
handleUploadCompletion(completed, failed){
   if (failed && failed.count){
      console.log(`${failed.count} files failed to upload`, failed.files) //or something
      retry(failed.files);
   }
   
   if (completed && completed.count){
      console.log(`${completed.count} files uploaded successfully`, completed.files )
   }
}
```

dev-KingMaster:
## Summary

- Add onAllUploadFinished to Uploader to notify when all in‑flight uploads complete.
- Closes #813.

## What’s New

- Prop: onAllUploadFinished?: (fileList: FileType[]) => void
- Fires once when pending uploads count drops to 0 (success, error, or manual cancel).
- Works with autoUpload and manual ref.current.start().

## Implementation

- Track in‑flight uploads with an internal counter; decrement on success/error/remove.
- Do not forward the prop to DOM nodes.
- Added PropTypes entry and unit test.

## Tests

- New test: src/Uploader/test/UploaderAllFinishedSpec.tsx
- Verified all Uploader tests pass locally.

## Usage
```
<Uploader action="/upload" onAllUploadFinished={files => toaster.push(<Message>All done</Message>)} />
```

## Backwards Compatibility

- Additive API, no breaking changes.

Contribution by Gittensor, learn more at https://gittensor.io/

### DIFF
--- a/src/Uploader/Uploader.tsx
+++ b/src/Uploader/Uploader.tsx
@@ -349,6 +349,30 @@ const Uploader = React.forwardRef((props: UploaderProps, ref) => {
     trigger.current?.clearInput();
   }, []);
 
+  // Decrement one pending entry for a batch and fire completion if empty
+  const finalizeOneInBatch = useCallback(
+    (batchId?: number) => {
+      if (typeof batchId !== 'number') return;
+      const map = pendingByBatchRef.current;
+      if (!map.has(batchId)) return;
+
+      const next = (map.get(batchId) || 0) - 1;
+      if (next < 0) {
+        if (process.env.NODE_ENV !== 'production') {
+          // eslint-disable-next-line no-console
+          console.warn('Uploader: pending count below zero for batch', batchId);
+        }
+      }
+      map.set(batchId, Math.max(0, next));
+      totalPendingRef.current = Math.max(0, totalPendingRef.current - 1);
+      if ((map.get(batchId) || 0) === 0) {
+        map.delete(batchId);
+        onAllUploadComplete?.(fileList.current);
+      }
+    },
+    [onAllUploadComplete, fileList]
+  );
+
   /**
    * Callback for successful file upload.
    * @param file
@@ -371,28 +395,9 @@ const Uploader = React.forwardRef((props: UploaderProps, ref) => {
       };
       updateFileStatus(nextFile);
       onSuccess?.(response, nextFile, event, xhr);
-      // Decrement pending counters for this batch
-      if (typeof batchId === 'number') {
-        const map = pendingByBatchRef.current;
-        if (map.has(batchId)) {
-          const next = (map.get(batchId) || 0) - 1;
-          if (next < 0) {
-            // Guard against logic regression in development
-            if (process.env.NODE_ENV !== 'production') {
-              // eslint-disable-next-line no-console
-              console.warn('Uploader: pending count below zero for batch', batchId);
-            }
-          }
-          map.set(batchId, Math.max(0, next));
-          totalPendingRef.current = Math.max(0, totalPendingRef.current - 1);
-          if ((map.get(batchId) || 0) === 0) {
-            map.delete(batchId);
-            onAllUploadComplete?.(fileList.current);
-          }
-        }
-      }
+      finalizeOneInBatch(batchId);
     },
-    [onSuccess, updateFileStatus, onAllUploadComplete, fileList]
+    [onSuccess, updateFileStatus, finalizeOneInBatch]
   );
 
   /**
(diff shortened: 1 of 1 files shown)


simonguo (changes requested):
### 1. Possible double or missed trigger
```js
if (pendingUploadsRef.current === 0 && anyStartedRef.current) {
  anyStartedRef.current = false;
  onAllUploadFinished?.(...);
}
```
If a new upload starts immediately after the previous batch finishes,
anyStartedRef.current might not yet be reset properly — leading to:
- the callback firing twice (for the old + new batch), or
- the callback being skipped for the new batch.

This happens because anyStartedRef and pendingUploadsRef aren’t tightly coupled to a single batch lifecycle.


### 2. Naming and Documentation

- `onAllUploadFinished` could be renamed to `onAllUploadComplete`, which is more natural and consistent with other callback styles.
- Update the documentation of Uploader, adding a description of the newly added `onAllUploadComplete`.

dev-KingMaster:
Hi @simonguo Thanks for your review.
- I’ve switched to batch‑scoped tracking to avoid double/missed triggers when a new batch starts as the previous one completes. Each started file is associated with a batch ID; the callback fires exactly once when that batch’s pending count reaches 0.
- Renamed the prop to onAllUploadComplete and updated types/PropTypes and tests accordingly. JSDoc now clarifies it fires once per batch.
- Also handled Copilot review as well

simonguo (changes requested):

simonguo (changes requested):
Review: onAllUploadComplete Implementation

Overall, the idea behind onAllUploadComplete is solid — it tracks upload completion per batch and fires once when all files in a batch have finished uploading.
However, there are several logical and edge-case issues in the current implementation that could cause incorrect or missed triggers.

### ✅ The Intended Behavior

The goal is clear:

- Each file selection or manual upload starts a new batch.
- Files in the same batch share a batchId.
- Once all files in that batch complete (success, error, or cancel), → trigger onAllUploadComplete(fileList).

This is a good design in theory.

### ⚠️ Problem 1: Batch Grouping Logic Is Not Reliable

Current logic:

```ts
let currentBatchId = batchIdCounterRef.current;
if (totalPendingRef.current === 0) {
  batchIdCounterRef.current += 1;
  pendingByBatchRef.current.set(batchIdCounterRef.current, 0);
  currentBatchId = batchIdCounterRef.current;
}
```

This means:

- A new batch is only created when there are no ongoing uploads.
- If the user adds more files while uploads are still pending, → those files are incorrectly grouped into the previous batch.

Example:

1.User selects 2 files → batch 1 starts.
2.While uploading, user selects 3 more files → still batch 1.
3.onAllUploadComplete won’t fire until all 5 files finish,
even though the user expected a new batch.

### ⚠️ Problem 2: Missing or Delayed finalizeOneInBatch Calls

The pending counter is decremented only when callbacks fire:

finalizeOneInBatch(batchId);

But:

- A file removed during upload may skip this call (even though you try to handle it).
- A network failure or component unmount could also bypass it.

Result:

The batch’s pending count can desync — leaving the map non-empty and `onAllUploadComplete` never firing.

### ⚠️ Problem 3: fileList.current May Not Reflect the Batch Snapshot

You call:

```ts
onAllUploadComplete?.(fileList.current);
```

At this point:

- The reducer may not yet have committed the latest file status.
- The user may have already added more files.

So the callback receives a global state, not the completed batch snapshot.

### ⚠️ Problem 4: Multiple Triggers Across Frequent Uploads

If users select files multiple times:

- Each batch may complete separately.
- onAllUploadComplete may fire multiple times in a short period.
- This may or may not be desired depending on intended UX.

### ✅ Recommended Fixes

Option 1 – Keep Batch Semantics (Recommended)

Create a new batch at the moment files are selected, not when uploading starts:

```ts
const batchId = ++batchIdCounterRef.current;
pendingByBatchRef.current.set(batchId, 0);

newFileList.forEach((file) => {
  fileKeyToBatchRef.current.set(file.fileKey, batchId);
});
```

Then pass this batchId into handleUploadFile.
Each selection = one batch, cleanly isolated.

### ➡️ Pros:

- Intuitive behavior (“each selection = one complete event”)
- Easy to track and debug

Option 2 – Drop Batch Concept, Use Global Completion

If the intention is simply:

“Fire once when all uploads (across all batches) are done.”

Then simplify:

```ts
if (totalPendingRef.current === 0) {
  onAllUploadComplete?.(fileList.current);
}
```

You can check this globally after each upload completion,
without tracking batches at all.

Issue #4359, opened 2025-07-31
Date Range Picker No Jump


edward-meister:
### What problem does this feature solve?

we are using single calendar in date range picker (https://rsuitejs.com/components/date-range-picker/#show-one-calendar), after start date was selected outside of current month, for selected end date picker jumps to today which causes distraction for user.

### What does the proposed API look like?

by passing a boolean to the date range picker, prevent it from going to today for end date.

Huyjoon1995:
facing the same issue now.

PR #4465, opened 2025-12-11
fix: date range picker not jumping to current day when selecting outside current calendar

Huyjoon1995:
### Description
Ref: https://github.com/rsuite/rsuite/issues/4359
we are using single calendar in date range picker (https://rsuitejs.com/components/date-range-picker/#show-one-calendar), after start date was selected outside of current month, for selected end date picker jumps to today which causes distraction for user. What i fix was just setting up the end date to be the same as the start date to allow user to select again after that to avoid this jump.
### Screenshot / Video

https://github.com/user-attachments/assets/1d986c71-a796-4b9a-82ae-f81a3e2c1ac3


Huyjoon1995:
hi @simonguo  i added the unit test. Thanks for the suggestion.

PR #4584, opened 2026-06-28
feat(locales): add Norwegian Nynorsk (nn_NO) locale

arvindfroi:
Follow-up to the Norwegian Bokmål (`nb_NO`) locale: adds Norwegian **Nynorsk** (`nn_NO`), the other official written standard of Norwegian.

- New `src/locales/nn_NO.ts`, mirroring the structure of `nb_NO`/`sv_SE` (uses the `nn` date-fns locale, Nynorsk date formats and weekday abbreviations, translated component strings).
- Registers `nnNO` in `src/locales/index.ts` (right after `nbNO`).

Disclosure: prepared with AI assistance and reviewed by me as a native Norwegian speaker.

### DIFF
--- a/src/locales/index.ts
+++ b/src/locales/index.ts
@@ -15,6 +15,7 @@ export { default as itIT } from './it_IT';
 export { default as kkKZ } from './kk_KZ';
 export { default as koKR } from './ko_KR';
 export { default as nbNO } from './nb_NO';
+export { default as nnNO } from './nn_NO';
 export { default as nlNL } from './nl_NL';
 export { default as ptBR } from './pt_BR';
 export { default as ruRU } from './ru_RU';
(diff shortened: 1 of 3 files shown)


simonguo (changes requested):
Hi @arvindfroi, thanks for adding the Nynorsk locale!

This PR currently has merge conflicts. The issue is that your branch was forked before **#4583** (nb_NO) was merged into `main`, so it still carries the nb_NO additions — which are now already in `main`. Here's what's conflicting:

- **`src/locales/index.ts`** — `nbNO` is already exported in `main`. You only need to add the `nnNO` line.
- **`src/locales/nb_NO.ts`** — this file already exists in `main` via #4583, so it shouldn't be part of this PR.

To fix:

```bash
git fetch upstream
git rebase upstream/main
```

During rebase, you should only keep the nn_NO.ts file and the nnNO export line in index.ts, and drop the nb_NO-related changes. After that, force-push and we'll get this merged.

